PR #1, opened 2016-12-01
Fix argument name

mvrilo:


progrium:
Last minute addition, thanks for catching that.

mvrilo:
:)

Issue #3, opened 2016-12-17
ssh.KeysEqual(nil, nil) panics


hloeffler:
ssh.KeysEqual() will panic if one of the "keys" is nil

i suggest to return false if one of the keys are nil
i suggest to return false if both keys are nil (because there are not keys)

PR #4, opened 2016-12-17
avoid panic at KeysEqual() if one of the keys is nil

hloeffler:
fixes https://github.com/gliderlabs/ssh/issues/3


progrium:
Great catch! Can you write a quick test for this? I know there aren't any tests yet, but they make sense for some of these functions.

hloeffler:
Thanks.
Sure, see last commit.

PR #15, opened 2017-02-03
Always respond to PTY requests

### LINKED ISSUE #13: Send failed reply when parsePtyRequest fails

belak:
Fixes #13


progrium:
Do you trust github.com/google/shlex? I used it a few years ago and was constantly adding runes to it and they haven't touched it. Working on Flynn we made our own fork but I think they stopped maintaining it, hence the library used now.

belak:
Ah, my mistake. That shouldn't have made it in.

PR #18, opened 2017-02-03
Ensure Session.Exit can only be called once

### LINKED ISSUE #17: Make it possible to send an exit-status other than 0
This will have to be a breaking change so we can get the return value from the handler.

EDIT: I've been informed this is already possible with Session.Exit, but there is an extra "exit-status" 0 being sent if that was called earlier in the connection.

belak:
Fixes #17 in an indirect way

### DIFF
--- a/session.go
+++ b/session.go
@@ -2,6 +2,7 @@ package ssh
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"net"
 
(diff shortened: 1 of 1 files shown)


belak:
If we're not worried about race conditions, this can just turn into a bool... I guess that would make more sense

progrium:
Luckily *we* handle sessions in a way that we can assume that, but it's *possible* the user could call Session.Exit in multiple goroutines, but I think worst case they get a different error.

Issue #22, opened 2017-02-06
Add a way to drain incoming connections


belak:
It sounds like `(*Server) Drain()` is what we're looking at, as well as adding an internal state on the Server so we can track running, draining, stopped.

Issue #20, opened 2017-02-04
Ensure SetOption is only callable if the server has not been started yet


belak:
Optionally, make a copy when calling makeConfig so the running server config can't be modified.

progrium:
Even better.

hloeffler:
+1 for copy, will help if multiple servers are used, with different config.
and please first add test for it

progrium:
Looks like we could still use a check if it's running in SetOption to bail if it's too late.

josegonzalez:
@progrium Do we want to use the `srv.state` idea @belak initially had in PR #23?

progrium:
Might be good to introduce.

PR #23, opened 2017-02-06
Add support for cleanly shutting down a server

belak:
This is related to #22 and #20 and is needed to write tests.

I understand this is probably more complicated than we were hoping, so feel free to offer suggestions for improvements.

Please do not merge this until I get a chance to double check stuff and clean it up.

### DIFF
--- a/server.go
+++ b/server.go
@@ -302,6 +302,18 @@ func (srv *Server) Shutdown(ctx context.Context) error {
 
 	select {
 	case <-ctx.Done():
+		srv.mu.Lock()
+		// If the state is still draining, kill everything so Serve will return
+		// cleanly.
+		if srv.state == stateDraining {
+			close(srv.killChan)
+		}
+		srv.mu.Unlock()
+
+		// If we've done our best to close connections, we should wait for all
+		// of them to clean up properly, even if there was a timeout because
+		// this should be fairly quick.
+		<-wgDoneChan
 	case <-wgDoneChan:
 	}
 
--- a/server.go
+++ b/server.go
@@ -243,6 +243,11 @@ func (srv *Server) Serve(l net.Listener) error {
 	}
 }
 
+// Close immediately closes the active net.Listener and any connections. For a
+// graceful shutdown, use Shutdown.
+//
+// Close returns any error returned from closing the Server's underlying
+// listener.
 func (srv *Server) Close() error {
 	srv.mu.Lock()
 
@@ -271,6 +276,11 @@ func (srv *Server) Close() error {
 	return lerr
 }
 
+// Shutdown gracefully shuts down the server without interrupting any active
+// connections. Shutdown works by first closing all open listeners then waiting
+// indefinitely for connections to return to idle and then shut down. If the
+// provided context expires before the shutdown is complete, then the all
+// connections are closed and the Context's error is returned.
 func (srv *Server) Shutdown(ctx context.Context) error {
 	srv.mu.Lock()
 
@@ -314,6 +324,8 @@ func (srv *Server) Shutdown(ctx context.Context) error {
 		// of them to clean up properly, even if there was a timeout because
 		// this should be fairly quick.
 		<-wgDoneChan
+
+		return ctx.Err()
 	case <-wgDoneChan:
 	}
 
--- a/server.go
+++ b/server.go
@@ -58,60 +58,6 @@ type Server struct {
 	killChan chan struct{}   // tracks if we're blindly killing connections
 }
 
-// getKillChan is used to get or create the kill channel
-func (srv *Server) getKillChan() chan struct{} {
-	srv.mu.Lock()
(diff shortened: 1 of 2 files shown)


belak:
Maybe something like srv.Shutdown(ctx) should be used (see https://tylerchr.blog/golang-18-whats-coming/)

server.go
```diff
@@ -21,6 +46,70 @@ type Server struct {
...
+	listener net.Listener    // listener currently being used by Serve, nil otherwise
+	mu       sync.Mutex      // general lock around the Server state
+	wg       *sync.WaitGroup // WaitGroup to track the number of connections still running
+	state    serverState     // tracks if the server is running or not
+	doneChan chan struct{}   // tracks if we want to exit
+	killChan chan struct{}   // tracks if we're blindly killing connections
+}
+
+// getKillChan is used to get or create the kill channel
+func (srv *Server) getKillChan() chan struct{} {
```
mark-adams: Why would we make this seem like magic instead of having an explicit initialization of the `killChan`? It seems like this could result in somebody not realizing there are side-effects here.

progrium: I generally agree, but I'm also a lot more tolerant when it comes to *internal* APIs for small projects.

belak: This is also similar to what `net/http` does. The locks and such are used more there because it's more complicated, but the general idea is the same. A large portion of this takes inspiration from there.

progrium: Oh, but also thanks for reviewing @mark-adams !

server.go
```diff
@@ -21,6 +46,70 @@ type Server struct {
...
+// closed. Note that this assumes srv.mu is held.
+func (srv *Server) lockedCloseDoneChan() {
+	if srv.doneChan != nil {
+		close(srv.doneChan)
+		srv.doneChan = nil
+	}
+}
+
+// getKillChan is used to get or create the wait group
+func (srv *Server) getWaitGroup() *sync.WaitGroup {
```
progrium: Like Mutex, a WaitGroup can be used as an empty value. Even the example does it this way: https://golang.org/pkg/sync/#example_WaitGroup

server.go
```diff
@@ -106,10 +255,93 @@ func (srv *Server) Serve(l net.Listener) error {
...
+
+	// Close the doneChan, killChan and listeners so we stop accepting new
+	// connections and ensure existing connections start getting shut down.
+	srv.lockedCloseDoneChan()
+	srv.lockedCloseKillChan()
+	lerr := srv.listener.Close()
+	srv.listener = nil
+
+	// Grab the waitgroup then unlock
+	wg := srv.wg
```
progrium: This shouldn't be necessary once changed to a value.

server.go
```diff
@@ -21,6 +46,70 @@ type Server struct {
...
+// lockedCloseKillChan will close the kill channel if it hasn't already been
+// closed. Note that this assumes srv.mu is held.
+func (srv *Server) lockedCloseKillChan() {
+	if srv.killChan != nil {
+		close(srv.killChan)
+		srv.killChan = nil
+	}
+}
+
+// getKillChan is used to get or create the done channel
```
progrium: Comment name discrepancy.

server.go
progrium: Now that I've read the full source, I think it does makes sense to *not* have this (or `getDoneChan`) and instead initialize in the one place these are called: when entering stateStarted in Serve().

server.go
```diff
@@ -106,10 +255,93 @@ func (srv *Server) Serve(l net.Listener) error {
...
+		return ErrInvalidState
+	}
+
+	// Close the listeners so everyone knows we're done.
+	srv.lockedCloseDoneChan()
+	lerr := srv.listener.Close()
+	srv.listener = nil
+
+	// Grab the waitgroup then unlock
+	wg := srv.wg
```
progrium: Same here.

server.go
belak: I've removed this helper function, but I still think it should be a pointer, otherwise there's a race condition between shutting down a server and starting a new one with the same Server object.

server.go
belak: All these helpers have been removed.

belak:
Once we agree on the general method, I'll add some tests.

belak:
Note to self: there's a small issue currently around the state getting set to stateClosing while the server is still draining.

belak:
There's a race condition somewhere (no surprise there) and the tests are failing sometimes.

`server_test.go:69: wait: remote command exited without exit status or exit signal`

Ideas would be appreciated.

progrium:
Shall we close in favor of https://github.com/gliderlabs/ssh/pull/34?

belak:
That's probably the best move. I'd like to get this in as it doesn't rely on polling to shut down, but the other method should work.

progrium:
Okay, well thanks for the contribution in any case and instigating a solution for this.

PR #25, opened 2017-02-15
fixed/finished basic pty support, added an example, and included tests

progrium:
cc @belak for review

Missing support for terminal modes, and still need signals, but omg tests.


progrium:
Last commit was a fix I added in working on a more complex example. You *always* end up writing the call to your "resizeWindow" function twice, once initially, once in the channel range. This PR changes behavior to where the window change channel has the initial window size in it, so you can just implement your range and that's it.

progrium:
Just added a new example that turns SSH into a "docker run" bridge ... in ~100 lines! Still no signals, but many containers work and I learned how to resolve some bugs in cmd.io, namely https://github.com/gliderlabs/cmd/issues/58 and https://github.com/gliderlabs/cmd/issues/59

belak:
I'm not sure how we want to handle this, but the `if req.WantReply` portions aren't strictly needed because it's handled in x/crypto/ssh:

```
func (r *Request) Reply(ok bool, payload []byte) error {
	if !r.WantReply {
		return nil
	}

	...
}
```

session.go
```diff
@@ -152,6 +152,9 @@ func (sess *session) handleRequests(reqs <-chan *gossh.Request) {
 			var kv = struct{ Key, Value string }{}
 			gossh.Unmarshal(req.Payload, &kv)
 			sess.env = append(sess.env, fmt.Sprintf("%s=%s", kv.Key, kv.Value))
+			if req.WantReply {
+				req.Reply(true, nil)
+			}
 		case "pty-req":
```
belak: How should this handle multiple pty-req messages before the exec/shell? Right now I think this will overwrite the pty, but if that's what we want, we need to be more careful about closing sess.winch because that could be double-closed.

session_test.go
```diff
@@ -0,0 +1,271 @@
...
+	"bytes"
+	"fmt"
+	"io"
+	"net"
+	"testing"
+
+	gossh "golang.org/x/crypto/ssh"
+)
+
+func (srv *Server) serveOnce(l net.Listener) error {
```
belak: Ah, this is a nice hack. I'll hopefully be able to cobble together something a bit more robust when the Shutdown pr goes in.

ssh.go
```diff
@@ -5,7 +5,6 @@ import (
 	"net"
 )
 
-// Signal as in RFC 4254 Section 6.10.
```
belak: Any reason this comment was removed?

progrium: I believe it's duplicated a few lines below.

session.go
progrium: Maybe the simplest solution is to ignore/deny (reply false) any more pty requests...

progrium:
@belak regarding  `if req.WantReply`, you're right. I know somebody removed them with a point, but I couldn't remember the point and added them back here. But I was wrong! Good catch.

session.go
progrium: Or perhaps once `pty` is set, it ignores the rest. I can imagine the callback denying for some reason and then it fixing and trying again. No idea if that's a realistic scenario with known SSH clients, but it makes sense implementation wise and resolves this issue.

PR #27, opened 2017-02-15
Add an experimental channel callback

belak:
This is a pretty small change which lets users deny ssh channels. We plan on using this to deny everything after the first one, but there may be other uses as well. It's modeled after the PtyCallback.


belak:
I'm going to close this for now since it sounds like most of the callbacks will be changing.

PR #29, opened 2017-02-22
contexts

progrium:
Implements a special ssh.Context that captures gossh.ConnMetadata and context.Context made for better callbacks and session setup, then exposed as regular context.Context on ssh.Session

Standardizes callbacks more while giving them access to more information and letting them contribute to state that would be made available to Session handler. Removes the need for PermissionsCallback as Permissions are exposed on ssh.Context. Generally simplifies code overall, while increasing capabilities, and remaining fairly Go idiomatic.

Todo: godocs, more tests

### DIFF
--- a/context.go
+++ b/context.go
@@ -14,26 +14,72 @@ type contextKey struct {
 }
 
 var (
-	ContextKeyUser          = &contextKey{"user"}
-	ContextKeySessionID     = &contextKey{"session-id"}
-	ContextKeyPermissions   = &contextKey{"permissions"}
+	// ContextKeyUser is a context key for use with Contexts in this package.
+	// The associated value will be of type string.
+	ContextKeyUser = &contextKey{"user"}
+
+	// ContextKeySessionID is a context key for use with Contexts in this package.
+	// The associated value will be of type []byte.
+	ContextKeySessionID = &contextKey{"session-id"}
+
+	// ContextKeyPermissions is a context key for use with Contexts in this package.
+	// The associated value will be of type *Permissions.
+	ContextKeyPermissions = &contextKey{"permissions"}
+
+	// ContextKeyClientVersion is a context key for use with Contexts in this package.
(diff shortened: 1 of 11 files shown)


progrium:
cc @belak

context.go
```diff
@@ -0,0 +1,102 @@
...
+
+// this is separate from newContext because we will get ConnMetadata
+// at different points so it needs to be applied separately
+func (ctx *sshContext) applyConnMetadata(conn gossh.ConnMetadata) {
+	if ctx.Value(ContextKeySessionID) != nil {
+		return
+	}
+	// for most of these, instead of converting to strings now, storing the byte
+	// slices means allocations only happen when accessing, not when contexts
+	// are being copied around
```
progrium: Though after I saw how context.WithValue is implemented this might not be an issue.

session.go
```diff
@@ -43,6 +44,10 @@ type Session interface {
 	// used it will return nil.
 	PublicKey() PublicKey
 
+	Context() context.Context
```
belak: Would it be worthwhile ensuring there are methods for everything exported by the ssh.Context? Especially if we're returning this as a context.Context here.

progrium: Sigh. The problem is the Session interface is already sooooo huuuuuge and it's only going to get bigger with signals and I also added Permissions. Interfaces should be small. A lot of the rest of what's in ssh.Context does qualify as "to inform, not to control" and is more for maintainers than users (at least how I interpret it). So it might feel inconsistent, but at the moment I'm ok with the tradeoff.

progrium:
Actually I think it's pretty much done with godocs and enough testing for now. Would like some more reviews first though.

progrium:
@shazow thanks!

hloeffler:
LGTM

progrium:
Merged after running in (alpha) production in Cmd.io for 20 days.

PR #31, opened 2017-03-15
agent forwarding support

progrium:
Originally we thought [crypto/ssh/agent](https://godoc.org/golang.org/x/crypto/ssh/agent) would give us this functionality, but that package is just the agent server and ssh client part. Basically the first leg. The second leg (between a forwarding unix socket made by the ssh server and the ssh client there) is not terribly well documented but is actually pretty straightforward.

Also included is an example. Documentation to come after review. Also, thoughts on testing this?

I also started the infrastructure for making channel type handlers easier to add. For some reason I was thinking I had to handle the agent forwarding channel, but it's actually done on the client side. So it was unnecessary, but we'll have to add new channel types eventually and this is a step in that direction.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -1,0 +1,70 @@
+package ssh
+
+import (
+	"io"
+	"io/ioutil"
+	"net"
+	"path"
+	"sync"
+
+	gossh "golang.org/x/crypto/ssh"
+)
+
+const (
+	agentRequestType = "auth-agent-req@openssh.com"
+	agentChannelType = "auth-agent@openssh.com"
+)
+
+var contextKeyAgentRequest = &contextKey{"auth-agent-req"}
+
+func setAgentRequested(sess *session) {
+	sess.ctx.SetValue(contextKeyAgentRequest, true)
+}
+
+func AgentRequested(sess Session) bool {
+	return sess.Context().Value(contextKeyAgentRequest) == true
+}
+
+func NewAgentListener() (net.Listener, error) {
+	dir, err := ioutil.TempDir("", "auth-agent")
+	if err != nil {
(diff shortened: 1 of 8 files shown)


agent.go
```diff
@@ -0,0 +1,70 @@
...
+			defer conn.Close()
+			channel, reqs, err := sshConn.OpenChannel(agentChannelType, nil)
+			if err != nil {
+				return
+			}
+			defer channel.Close()
+			go gossh.DiscardRequests(reqs)
+			var wg sync.WaitGroup
+			wg.Add(2)
+			go func() {
```
shazow: Some comments wouldn't hurt, like why we're only closing the Write side.

agent.go
```diff
@@ -0,0 +1,70 @@
...
+func AgentRequested(sess Session) bool {
+	return sess.Context().Value(contextKeyAgentRequest) == true
+}
+
+func NewAgentListener() (net.Listener, error) {
+	dir, err := ioutil.TempDir("", "auth-agent")
+	if err != nil {
+		return nil, err
+	}
+	l, err := net.Listen("unix", path.Join(dir, "listener.sock"))
```
shazow: Prolly worth shoving "listener.sock" in a const.

progrium: 👍

agent.go
progrium: Oh I should double check this. I copied it from the agent package which is handling the other end.

progrium: both objects get a close when the function returns, but closing write is definitely necessary

progrium:
Going to merge as we've been running in cmd.io for a bit now.

PR #34, opened 2017-04-15
server: first pass at Shutdown and Close

mattaitchison:
Most of the code is an exact copy of [net/http/server.go](https://github.com/golang/go/blob/master/src/net/http/server.go).

closes: #22

### DIFF
--- a/server.go
+++ b/server.go
@@ -13,7 +13,7 @@ import (
 
 // ErrServerClosed is returned by the Server's Serve, ListenAndServe,
 // and ListenAndServeTLS methods after a call to Shutdown or Close.
-var ErrServerClosed = errors.New("http: Server closed")
+var ErrServerClosed = errors.New("ssh: Server closed")
 
 // Server defines parameters for running an SSH server. The zero value for
 // Server is a valid configuration. When both PasswordHandler and
--- a/server.go
+++ b/server.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"net"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	gossh "golang.org/x/crypto/ssh"
@@ -31,11 +30,10 @@ type Server struct {
 
 	channelHandlers map[string]channelHandler
 
-	mu         sync.Mutex
-	inShutdown int32 // accessed atomically (non-zero means we're in Shutdown)
-	listeners  map[net.Listener]struct{}
-	activeConn map[*gossh.ServerConn]struct{}
-	doneChan   chan struct{}
+	mu        sync.Mutex
+	listeners map[net.Listener]struct{}
+	conns     map[*gossh.ServerConn]struct{}
+	doneChan  chan struct{}
 }
 
 // internal for now
@@ -103,9 +101,9 @@ func (srv *Server) Close() error {
 	defer srv.mu.Unlock()
 	srv.closeDoneChanLocked()
 	err := srv.closeListenersLocked()
-	for c := range srv.activeConn {
+	for c := range srv.conns {
 		c.Close()
-		delete(srv.activeConn, c)
+		delete(srv.conns, c)
 	}
 	return err
 }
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -1,13 +1,21 @@
...
+	"sync"
+	"sync/atomic"
 	"time"
 
 	gossh "golang.org/x/crypto/ssh"
 )
 
+// ErrServerClosed is returned by the Server's Serve, ListenAndServe,
+// and ListenAndServeTLS methods after a call to Shutdown or Close.
+var ErrServerClosed = errors.New("http: Server closed")
```
mattaitchison: Remove references to http

server.go
```diff
@@ -79,6 +93,67 @@ func (srv *Server) Handle(fn Handler) {
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+		}
+	}
+
+}
+
+func (s *Server) shuttingDown() bool {
```
mattaitchison: Remove unused function

server_test.go
```diff
@@ -0,0 +1,102 @@
...
+		err := s.Close()
+		if err != nil {
+			t.Fatal(err)
+		}
+	}()
+
+	timeout := time.After(time.Millisecond * 100)
+	select {
+	case <-timeout:
+		t.Error("timeout")
```
mattaitchison: Change to fatal?

server.go
```diff
@@ -79,6 +93,67 @@ func (srv *Server) Handle(fn Handler) {
...
+	srv.closeDoneChanLocked()
+	srv.mu.Unlock()
+	ticker := time.NewTicker(shutdownPollInterval)
+	defer ticker.Stop()
+	for {
+
+		srv.mu.Lock()
+		activeConns := len(srv.activeConn)
+		srv.mu.Unlock()
+		if activeConns == 0 {
```
mattaitchison: Refactor activeConn check into function?

server.go
```diff
@@ -79,6 +93,67 @@ func (srv *Server) Handle(fn Handler) {
...
+	srv.mu.Lock()
+	lnerr := srv.closeListenersLocked()
+	srv.closeDoneChanLocked()
+	srv.mu.Unlock()
+	ticker := time.NewTicker(shutdownPollInterval)
+	defer ticker.Stop()
+	for {
+
+		srv.mu.Lock()
+		activeConns := len(srv.activeConn)
```
progrium: Since there are no idle connections, maybe we just call them conns?

server_test.go
progrium: Why or why not

progrium:
@shazow quick review?

server.go
```diff
@@ -79,6 +93,67 @@ func (srv *Server) Handle(fn Handler) {
...
+// exercise for the reader.
+var shutdownPollInterval = 500 * time.Millisecond
+
+// Shutdown gracefully shuts down the server without interrupting any
+// active connections. Shutdown works by first closing all open
+// listeners, and then waiting indefinitely for connections to close.
+// If the provided context expires before the shutdown is complete,
+// then the context's error is returned.
+func (srv *Server) Shutdown(ctx context.Context) error {
+	atomic.AddInt32(&srv.inShutdown, 1)
```
shazow: Any reason to do this atomically rather than inside of the mu mutex?

(Also looks like inShutdown is not needed anyways?)

server.go
```diff
@@ -79,6 +93,67 @@ func (srv *Server) Handle(fn Handler) {
...
+		c.Close()
+		delete(srv.activeConn, c)
+	}
+	return err
+}
+
+// shutdownPollInterval is how often we poll for quiescence
+// during Server.Shutdown. This is lower during tests, to
+// speed up tests.
+// Ideally we could find a solution that doesn't involve polling,
```
shazow: It's not clear to me why polling is necessary? Are we waiting for all of the clients to successfully disconnect before Close() returns? Is this a requirement?

In my code, I typically disable the server from accepting new connections and then in separate goroutine(s) I start issuing disconnect orders/closing connections while the original goroutine just returns.

PR #36, opened 2017-04-20
ConnCallback

progrium:
This is a relatively quick pass at a new callback that lets you wrap the underlying connection used by SSH, as well as do any other initial connection validation before processing, like determining connection should be dropped for rate limiting.

It also has a `WrapConn` option helper, just because it seems consistent, but it makes me wonder, do we need a helper for every field of server configuration?


progrium:
Also yes, the test has a race. Not entirely sure how to do that test without clean shutdown.

progrium:
Need to move forward with this. Does this seem ok?

server.go
```diff
@@ -27,6 +27,7 @@ type Server struct {
 	PasswordHandler             PasswordHandler             // password authentication handler
 	PublicKeyHandler            PublicKeyHandler            // public key authentication handler
 	PtyCallback                 PtyCallback                 // callback for allowing PTY sessions, allows all if nil
+  ConnCallback                ConnCallback                // optional callback for wrapping net.Conn before handling
```
belak: Just noticed this... looks like it's indented with spaces? A gofmt might be prudent.

progrium:
Any suggestions to fix the data race in the test?

belak:
You could send "written" on a buffered channel and do a read on that channel with a timeout... then you know it won't hang, but it shouldn't be a race either. Alternatively, `atomic.Value` may be an option.

progrium:
Going to add an example using this and then merge.

PR #38, opened 2017-04-27
Support for local port forwarding

notnoopci:
Permit use of local port forwarding (e.g. `ssh -L 8080:example.com:80 ssh-server`).

Due to our needs (at CircleCI) so far, I'm restricting this to local port forwarding.  If it matches conventions or pattern, we can contribute remote port forwarding and SOCKS as well in a follow up PR.

Few considerations made to call out:
* Defaults to disallowing port forwarding.  With the library being used in contexts without providing shell, defaulting to allowing it exposes a security vector to existing library consumers.  It also feels like the right default regardless IMO
* I used verbose `LocalPortForwarding` name in the public API thinking it matches the high level nature of the library.  Open for ideas there.
* Not sure what the testing pattern here is - so could use feedback

### DIFF
--- a/ssh.go
+++ b/ssh.go
@@ -42,6 +42,9 @@ type PasswordHandler func(ctx Context, password string) bool
 // PtyCallback is a hook for allowing PTY sessions.
 type PtyCallback func(ctx Context, pty Pty) bool
 
+// LocalPortForwardingCallback is a hook for allowing port forwarding
+type LocalPortForwardingCallback func(ctx Context, destinationHost string, destinationPort uint32) bool
+
 // Window represents the size of a PTY window.
 type Window struct {
 	Width  int
--- a/tcpip.go
+++ b/tcpip.go
@@ -1,0 +1,58 @@
+package ssh
+
+import (
+	"fmt"
+	"io"
+	"net"
+
+	gossh "golang.org/x/crypto/ssh"
+)
+
+// direct-tcpip data struct as specified in RFC4254, Section 7.2
+type forwardData struct {
+	DestinationHost string
+	DestinationPort uint32
+
+	OriginatorHost string
+	OriginatorPort uint32
+}
+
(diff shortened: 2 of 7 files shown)


ssh.go
```diff
@@ -42,6 +42,9 @@ type PasswordHandler func(ctx Context, password string) bool
 // PtyCallback is a hook for allowing PTY sessions.
 type PtyCallback func(ctx Context, pty Pty) bool
 
+// LocalPortForwardingCallback is a hook for allowing port forwarding
+type LocalPortForwardingCallback func(ctx Context, destinationHost string, destinationPort uint32) bool
```
notnoopci: I chose plain parameters here rather than a typed struct, mostly because I'm unhappy with the names - and it's easier to change parameter names than struct field names and I don't envision more parameters so far.  Happy to change it if I get definitive names.

tcpip.go
```diff
@@ -0,0 +1,58 @@
...
+		newChan.Reject(gossh.Prohibited, "port forwarding is disabled")
+		return
+	}
+
+	dest := fmt.Sprintf("%s:%d", d.DestinationHost, d.DestinationPort)
+
+	var dialer net.Dialer
+	dconn, err := dialer.DialContext(ctx, "tcp", dest)
+	if err != nil {
+		newChan.Reject(gossh.ConnectionFailed, err.Error())
```
notnoopci: The rejects reasons and messages are are approximate and don't match the behavior of OpenSSH server - don't see expected rejection reasons in the spec for these cases.

ssh.go
progrium: How are you unhappy with the names? Are these what are used in the protocol spec?

progrium:
Overall this looks great. Simple and follows our conventions so far. I'd love to see some tests. Also we'd love to see remote forwarding and SOCKS, but keep in mind we'll probably change the public API a bit afterwards to simplify and would love to have your feedback in that process.

notnoopci:
Updated with tests!  Thanks!

Also, expect an attribution notice for CircleCI 2.0 very soon!

PR #39, opened 2017-04-29
update config for CircleCI 2.0

notnoopci:
Update repo to use 2.0.  Build time is [7 seconds](https://circleci.com/gh/notnoopci/ssh/4) compared to [1.5 minutes](https://circleci.com/gh/gliderlabs/ssh/21).

A bit self-serving I must admit - but noticed that other gliderlabs public repos already use 2.0 - so felt more OK doing it :).

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,3 +1,10 @@
-test:
-  override:
-    - go test -v -race
+version: 2.0
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


progrium:
Thanks! We would have gotten to it eventually because we love 2.0.

PR #41, opened 2017-05-22
[+] Handle client signals

antmanler:
A lightweight observable list to support multiple subscriber to handle signal

### DIFF
--- a/session.go
+++ b/session.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"sync"
 
 	"github.com/anmitsu/go-shlex"
 	gossh "golang.org/x/crypto/ssh"
@@ -57,7 +58,8 @@ type Session interface {
 	// of whether or not a PTY was accepted for this session.
 	Pty() (Pty, <-chan Window, bool)
 
-	// TODO: Signals(c chan<- Signal)
+	// Signals returns a SingalChan to handle clients' signals
+	Signals() SignalChan
 }
 
 func sessionHandler(srv *Server, conn *gossh.ServerConn, newChan gossh.NewChannel, ctx *sshContext) {
@@ -73,6 +75,9 @@ func sessionHandler(srv *Server, conn *gossh.ServerConn, newChan gossh.NewChanne
 		ptyCb:   srv.PtyCallback,
 		ctx:     ctx,
 	}
+
+	// create empty signal stream
+	sess.signal.stream = newSignals("")
 	sess.handleRequests(reqs)
 }
 
@@ -88,6 +93,11 @@ type session struct {
 	ptyCb   PtyCallback
 	cmd     []string
 	ctx     *sshContext
+
+	signal struct {
+		sync.RWMutex
+		stream *signals
+	}
 }
 
 func (sess *session) Write(p []byte) (n int, err error) {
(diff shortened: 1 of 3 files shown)


progrium:
Hi @antmanler, I'm excited to get this feature added, but your implementation is much more complicated than I would have expected. What was the reason to have more than just a method that takes a channel of Signals, similar to the standard library for handling signals?

antmanler:
@progrium firstly, sorry for my poor english :)

Checking standard library and your [original comments](https://github.com/gliderlabs/ssh/blob/master/session.go#L60) in source, handling signal should be a pub/sub like interface(one publisher broadcast to many subscribers).

There are two patterns to implement such interface

```
Subscribe(chan<-)  // go standard library
Subscribe() subscription
```

**problems using channel**

In go it's hard to implement broadcasting using go channel (ref to standard `signal.Notify`)

keeping a list(or map) of subscribers, for each received signal forwards to downstream

```go
// first approach, stupid simple
for _, c := subscribers {
    c <- signal // this may block, if c is full
}

// second approach, signal.Notify style
for _, c := subscribers {
    select {
    case c <- signal:
    default:    // this will miss slow subscriber
    }
}
```

if let user provide a channel or callback, we have to support how to unsub or stop sending to such channel

**broadcasting utilize channel**

Channel in go is not support broadcasting, except closing a unbuffered channel

using a linked list

```go
type pubChan struct {
    notifyCh chan struct{}
    payload  Something  // the state(value) causing notify channel closed

    next *pubChan
}

type app struct {
    //...
    reactField pubChan
}

```

session keeps latest `pubChan` as `reactField`,  when comes a message(signal) append a `pubChan` to `next`, and closes `notifyCh` to broadcast to all subscribers, then replace the `reactField` with `reactField.next`

**benefits of linked list approach**

* implement the broadcasting simply
* do not block
* do not need to maintain list of subscribers
* the speed for consuming messages is independent to each subscriber

**caveats**

* if a subscriber hold a old `pubChan` and forget to consume it will produce a long linked list in memory that will not be collected by GC (mostly this should be a bug in app)

progrium:
I agree this approach is better for pubsub, but I don't believe the common case is a one to many broadcast. It will most often be one to one. If the user has a more complex scenario, they can wrap it to provide your API. The code you have is "simple" but it's still much more than I'd expect for the value of the feature. If that makes sense. 

I'm not even sure unsubscribing is necessary. I know that sounds crazy, but like PTY, if you're going to support PTY, you are going to handle PTY window resizes until the session is over. Similarly, it's very unlikely if you want to handle signals that you are going to want to stop before the session is over. 

Anybody else have thoughts?

antmanler:
I agree it's a little bit complicated for user, they have to learn a new concept, and a api should keep simple to use, but the cohesion should be considered as well.  I think we should reduce the impact of user's misuse, if we provides a API like the following
```go
Notify(c chan<- Signal)
```
this is the go `signal.Notify` style which implies to users that they can call it multiple times with different channels

or `time.Ticker` style:
```go
C() <-chan Signal
```
this may be better, the user knows it's a read only channel that do not support multiple subscribers

But, neither of those approaches solves the situation that user may cause a channel leak and blocking the logic of handling gossh.Request.

I feel so hard to decide which is better, maybe a simple interface with detailed documents and examples and accept the fact that user can break our code :(

belak:
We could just try to match the api for `signal.Notify`. It specifically doesn't block ("Package signal will not block sending to c: the caller must ensure that c has sufficient buffer space to keep up with the expected signal rate. For a channel used for notification of just one signal value, a buffer of size 1 is sufficient.")... It could be done (for now) with the caveat that we only accept one call to it per Session channel which should keep the code fairly simple.

progrium:
Or even simply say every call to it replaces the channel used in the previous call if any. And yes, I learned early on you should generally let the user pass the channel because that lets them decide if it's buffered or not.

progrium:
Closing in favor of simpler #44, which is loosely based on this PR.

PR #44, opened 2017-06-30
session: adding signal handling support

progrium:
A simpler implementation for signal handling based on the work done in https://github.com/gliderlabs/ssh/pull/41. It also buffers signals since signals may be sent before a channel is registered. The API is very similar to the `signal` package, but only allowing one channel for all signals, and unregistering by calling again with `nil`.

### DIFF
--- a/session.go
+++ b/session.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"sync"
 
 	"github.com/anmitsu/go-shlex"
 	gossh "golang.org/x/crypto/ssh"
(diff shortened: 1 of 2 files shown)


session.go
```diff
@@ -57,9 +58,19 @@ type Session interface {
...
 	Pty() (Pty, <-chan Window, bool)
 
-	// TODO: Signals(c chan<- Signal)
+	// Signals registers a channel to receive signals sent from the client. The
+	// channel must handle signal sends or it will block the SSH request loop.
+	// Registering nil will unregister the channel from signal sends. During the
+	// time no channel is registered signals are buffered up to a reasonable amount.
+	// If there are buffered signals when a channel is registered, they will be
+	// sent in order on the channel immediately after registering.
+	Signals(c chan<- Signal)
```
belak: This is a simpler version, but it doesn't allow for everything that the stdlib `signal.Notify` does (in particular adding multiple channels and stopping them independently).

progrium: Personally I really don't see why we should. It's a lot more management of something that isn't even used that much. OpenSSH doesn't even handle signals. I only today realized that most signals via PTY are sent via escape codes, so the majority of the common signal cases are already handled.

progrium:
Anybody 👍 ?

belak:
Ok, I've looked over this code really quickly... I've got a few comments:

* As mentioned before, I think it may make sense to allow more than one callback, but this isn't super important, especially because it makes removing them harder - something signal.Notify doesn't allow.
* The sigBuf seems unnecessary. It may be easier to do what signal.Notify does and just drop signals if there's noone waiting to receive them. This relies on the user to provide a buffered chan and/or handle them as soon as they come in, rather than moving that work into this library. If you select on a channel send and default, it should provide this.

progrium:
After finding out OpenSSH doesn't implement this I'm even less concerned that you can't register multiple channels. If you need multiple channels for some reason you can still make it work.

I'd like sigBuf to be unnecessary but signals can be sent before the handler is even run. This is unlikely but since the main use case for these signals is automated signals, it's possible signals could be sent immediately. For example, in the test, which will fail without it for this reason. Unless a sleep is introduced. :\

progrium:
Objections if I merge?

belak:
This looks fine to me, but it's usefulness seems a bit limited if it isn't even supported by OpenSSH... and could confuse the user on that point.

Issue #11, opened 2017-02-03
Allow setting timeouts for connection


belak:
Specifically for the handshake, requesting a session, starting a command, and connection closing.

progrium:
Do you have any example of libraries that let you specify all the possible timeouts? Usually there is a high level timeout or two, but SSH has a lot of ones as you mention. Perhaps a `Timeout` struct than can be passed into our server object.

belak:
I don't have any public examples. This comes from our config struct though (all the ones we support internally):

```go
	// HandshakeTimeout is the amount of time allowed between connecting and
	// successful auth.
	HandshakeTimeout time.Duration

	// NewChannelTimeout is the amount of time allowed between auth and making
	// the first valid channel.
	NewChannelTimeout time.Duration

	// ChannelRequestTimeout is the amount of time allowed between creating a
	// channel and making the first valid request.
	ChannelRequestTimeout time.Duration

	// ConnectionCloseTimeout is the amount of time allowed between when the
	// command finishes running and the connection is forcefully closed.
	ConnectionCloseTimeout time.Duration

```

belak:
I realize this won't be used by everyone, or even most people, so I'd be happy to do the work to support most of the features I'm proposing.

progrium:
What do you think of wrapping these in a struct:
```
type Timeouts struct {
	// HandshakeTimeout is the amount of time allowed between connecting and
	// successful auth.
	Handshake time.Duration

	// NewChannelTimeout is the amount of time allowed between auth and making
	// the first valid channel.
	NewChannel time.Duration

	// ChannelRequestTimeout is the amount of time allowed between creating a
	// channel and making the first valid request.
	ChannelRequest time.Duration

	// ConnectionCloseTimeout is the amount of time allowed between when the
	// command finishes running and the connection is forcefully closed.
	ConnectionClose time.Duration
}
```

belak:
That works for me. Other notes from our discussion:

* ConnectionClose is really only useful if #7 is also done
* It may also be useful to have a ChannelClose timeout

aderouineau:
I have created a pull request to support handshake timeout: #204

bavarianbidi:
I've stumbled over this issue here and i'm wondering if it's also a valid use-case to have a timeout where the `server` will shutdown if no connection is established within the defined timeout duration?

more context:

we are using https://github.com/owenthereal/upterm which spawn a `sshServer` (in `gliderlabs/ssh` speech) on demand.
[...]

PR #46, opened 2017-07-14
server: timeouts and context canceling on closed connections

### LINKED ISSUE #45: Cancel the Session.Conetxt() when client connection closes
**Reference**: https://godoc.org/net/http#Request.Context

> For outgoing client requests, the context controls cancelation.
>
> For incoming server requests, the context is canceled when the client's connection closes, the > request is canceled (with HTTP/2), or when the ServeHTTP method returns.

This is useful for using `Session.Conetxt()` to do something in background.

progrium:
Closes #45. Also adds timeouts, but not nearly as many as some may want (#11).

Timeouts example updated to just use builtin timeouts and also shows using Context for detecting closed connections.

Signed-off-by: Jeff Lindsay <progrium@gmail.com>

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1,0 +1,50 @@
+package ssh
+
+import (
+	"context"
+	"net"
+	"time"
+)
+
+type serverConn struct {
+	net.Conn
+
+	idleTimeout   time.Duration
+	maxDeadline   time.Time
+	closeCanceler context.CancelFunc
+}
+
+func (c *serverConn) Write(p []byte) (n int, err error) {
+	c.updateDeadline()
+	n, err = c.Conn.Write(p)
+	if _, isNetErr := err.(net.Error); isNetErr && c.closeCanceler != nil {
+		c.closeCanceler()
+	}
+	return
+}
+
+func (c *serverConn) Read(b []byte) (n int, err error) {
+	c.updateDeadline()
+	n, err = c.Conn.Read(b)
+	if _, isNetErr := err.(net.Error); isNetErr && c.closeCanceler != nil {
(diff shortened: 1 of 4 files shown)


server.go
```diff
@@ -191,17 +194,25 @@ func (srv *Server) Serve(l net.Listener) error {
...
 			return
 		}
-		conn = cbConn
+		newConn = cbConn
+	}
+	ctx, cancel := newContext(srv)
+	conn := &serverConn{
+		Conn:          newConn,
+		idleTimeout:   srv.IdleTimeout,
+		closeCanceler: cancel,
```
tsl0922: Maybe `defer cancel()` here is more clean?

progrium: The point was to give the connection the `cancel` not just for when it's closed but when `net.Error` is returned from a Read or Write.

conn.go
```diff
@@ -0,0 +1,50 @@
...
+	err = c.Conn.Close()
+	if c.closeCanceler != nil {
+		c.closeCanceler()
+	}
+	return
+}
+
+func (c *serverConn) updateDeadline() {
+	idleDeadline := time.Now().Add(c.idleTimeout)
+	if idleDeadline.Unix() < c.maxDeadline.Unix() {
```
nhooyr: There is a bug on this line. It's possible that the idleTimeout is the zero value but maxDeadline is not. So what will happen is that the line below will always set the deadline to the current time and an error will occur.

progrium: Great catch. It's already merged though, so maybe you can submit a PR?

nhooyr: will do

PR #48, opened 2017-07-20
Added LocalAddr() to session interface

Mester19:


progrium:
I guess I can't complain about this.

PR #54, opened 2017-08-26
Updated SSH docker example

stevemurr:
Updated function signature in ssh-docker example.  Included a Dockerfile and readme/tutorial.


progrium:
Readme could probably be cleaned up more in terms of formatting but it's good enougghhhh.

PR #59, opened 2017-10-31
Move channelHandlers to avoid data race

phil-halley:
There was a data race on the channelHandlers variable when there were multiple simultaneous connections.

Moving this function as in this PR is one option to resolve, and prepares for future changes if you wanted channelHandlers to be overridable. 

Another option would just be to add srv.mu.Lock() and Unlock() around the current location of the code in the config function and the read in handleConn, but the changes in this PR seemed cleaner to me.

### DIFF
--- a/server.go
+++ b/server.go
@@ -56,10 +56,6 @@ func (srv *Server) ensureHostSigner() error {
 }
 
 func (srv *Server) config(ctx *sshContext) *gossh.ServerConfig {
-	srv.channelHandlers = map[string]channelHandler{
-		"session":      sessionHandler,
-		"direct-tcpip": directTcpipHandler,
-	}
 	config := &gossh.ServerConfig{}
 	for _, signer := range srv.HostSigners {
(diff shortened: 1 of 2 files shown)


progrium:
Thanks! Actually, can you take a look at this PR I just opened where I believe I also fixed this. A review of the whole thing would be great, but if you just want to see if it solves this problem that's fine too. :D 

https://github.com/gliderlabs/ssh/pull/60

progrium:
Hey Phil, since my PR is taking so long, I'll merge this in if you can resolve the conflict with latest master.

phil-halley:
I've updated this PR to resolve the merge conflict. Thanks!

#60 also looks like it would address this, and the locking there looks like the right approach. I'm afraid I don't know enough about ssh to be comfortable to  offer a full review on $60.

e3b0c442:
Hello,

Any blockers on merging this or additional work that needs to be done?

progrium:
We shall see now!

PR #60, opened 2017-10-31
Remote forwarding

progrium:


belak:
I'm not very familiar with how tcp forwarding works over SSH, so I'm not sure if I'm the best person to review this. I'm also not sure when I'll be able to give it the time it deserves.

On a quick glance, it doesn't look like there's any way to disable tcp forwarding? It seems like that would be an essential option so people can avoid users of the ssh service being able to tunnel into an internal network (along with other potential security concerns).

progrium:
Ah. Yeah an option that adds it to the handlers would be nice. Any review of the Go code would be nice though, too. If you have time!

tcpip.go
```diff
@@ -56,3 +66,125 @@ func directTcpipHandler(srv *Server, conn *gossh.ServerConn, newChan gossh.NewCh
...
+			return false, []byte{}
+		}
+		addr := net.JoinHostPort(reqPayload.BindAddr, strconv.Itoa(int(reqPayload.BindPort)))
+		ln, err := net.Listen("tcp", addr)
+		if err != nil {
+			// TODO: log listen failure
+			return false, []byte{}
+		}
+		_, destPortStr, _ := net.SplitHostPort(ln.Addr().String())
+		destPort, _ := strconv.Atoi(destPortStr)
```
shazow: Skipping confirmation here feels risky, you might end up with :0 which will bind to a random port (?) which is unexpected?

_examples/ssh-remoteforward/remoteforward.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"io"
+	"log"
+
+	"github.com/gliderlabs/ssh"
+)
+
+func main() {
+	ssh.Handle(func(s ssh.Session) {
+		io.WriteString(s, "Remote forwarding available...\n")
```
shazow: Does this mean that forwarding is on by default? (Is that a good idea?)

progrium: Right, prob not. Now see https://github.com/gliderlabs/ssh/issues/68

tcpip.go
progrium: Ah, you mean error checking.

eliastor:
Are there any movements planned for this PR?

progrium:
No plans from me in immediate future.

Issue #56, opened 2017-09-18
HostKeyFile Not support ED25519


fcharlie:
When I use `ssh.HostKeyFile` add a `/etc/ssh/ssh_host_ed25519_key`, Start ssh server error, Report: `unsupported key type 'OPENSSH PRIVATE KEY'`

In fact, when i use `ssh.ParsePrivateKey`(x/crypto/ssh),  it can support **ed25519**
```go
func AddHostKeyFile(filepath string, srv *ssh.Server) {
	pemBytes, err := ioutil.ReadFile(filepath)
	if err != nil {
		log.Printf("AddHostKeyFile: %s", err)
		return
	}
	hostKey, err := gossh.ParsePrivateKey(pemBytes)
	if err != nil {
		log.Printf("Fatal to parse host key: %s, %s", filepath, err)
		return
	}
	srv.AddHostKey(hostKey)
	return
}
```

Bug in: 
https://github.com/gliderlabs/ssh/blob/4a4de396c4d49400d09e0b4d83053a01539e4bcc/util.go#L14

```go
func signerFromBlock(block *pem.Block) (ssh.Signer, error) {
	var key interface{}
	var err error
	switch block.Type {
	case "RSA PRIVATE KEY":
		key, err = x509.ParsePKCS1PrivateKey(block.Bytes)
	case "EC PRIVATE KEY":
		key, err = x509.ParseECPrivateKey(block.Bytes)
	case "DSA PRIVATE KEY":
		key, err = ssh.ParseDSAPrivateKey(block.Bytes)
	default:// NOT parse ed25519
		return nil, fmt.Errorf("unsupported key type %q", block.Type)
	}
	if err != nil {
		return nil, err
	}
	signer, err := ssh.NewSignerFromKey(key)
	if err != nil {
		return nil, err
	}
	return signer, nil
}
```

progrium:
Yep, easy fix is to use `ssh.ParsePrivateKey` from `x/crypto/ssh`. I don't believe it existed at the time. Would love a PR.

belak:
It looks like signerFromBlock is pretty close to a direct copy from [ParseRawPrivateKey](https://github.com/golang/crypto/blob/master/ssh/keys.go#L806). The main downside to ssh.ParsePrivateKey is that it only supports a single private key block, while the code here supports adding multiple keys from a single file.

progrium:
[...]

PR #61, opened 2017-11-01
Switch to using gossh.ParsePrivateKey when reading private keys

belak:
As mentioned in #56 this means we would only be able to parse one private key per file, but I don't think it's common practice to store multiple private ssh keys in one file.


belak:
Note that this isn't currently hit by any tests and I haven't had the chance to use this extensively yet.

PR #64, opened 2017-11-21
Support sftp

soopsio:


progrium:
Looks like a good start! Do you think sftp needs to be hard coded like that?

soopsio:
I think sftp and env and shell are the same type, I know the subsystem only sftp, so hard coded.

PR #65, opened 2017-11-21
Support sftp subsystem

soopsio:


session.go
```diff
@@ -64,6 +64,9 @@ type Session interface {
 	// of whether or not a PTY was accepted for this session.
 	Pty() (Pty, <-chan Window, bool)
 
+	// Sftp returns a boolean, when subsystem is sftp, return true
+	Sftp() bool
```
progrium: Maybe this should be `Subsystem(string) bool` so it can be used for other subsystems.

soopsio: I have already modified,please review it, thank you.

PR #66, opened 2017-11-22
Add support for subsystem requests

ucirello:


progrium:
Is this an alternative to the other PR for subsystems? Can you elaborate on this a bit?

ucirello:
> Is this an alternative to the other PR for subsystems? 

Actually, this was a coincidence. I was doing something on my own, changed this library and opened the PR. Only later that I saw that there was another PR kind of approaching a similar problem.

I believe this solution is a superset of #65. 

> Can you elaborate on this a bit?

The idea here is to give a general interface for `subsystem` calls - not only `sftp`. It turns out that with a very small change in `Session`, it is possible to serve all `subsystem` calls. 

Also, this differs from the other proposal by ensuring `subsystem` handlers are not mixed up with `exec` and `shell` handlers. Although the exposed network is the same, `subsystem` calls usually need some magic cookie parsing to actually decide what to serve, hence the addition of `Payload() []byte` call instead of testing for subsystem type in `req.Type` switch-block.

It is debatable whether the subsystem matching could be made at the library level. This would make this library easier to use, but harder to predict its performance (assuming multiple subsystem handlers with automatic matching). Thus I preferred to make something simpler and easier to grok.

ucirello:
ping @progrium

progrium:
This would progress faster if there weren't 2 PRs for it. I just have to review them. If anybody has input that would help, that could move things along.

ucirello:
@progrium I understand the predicament. I am closing this PR in the favor of the other one. If the other isn't merged, then I will reopen this.

PR #69, opened 2017-12-28
small api refactor

moul:
These updates make it easier to implement and pass custom Session and
Context implementations

No compatibility breaking, all tests pass


progrium:
This is great, thanks!

PR #74, opened 2018-04-17
Update shutdown to use a WaitGroup rather than sleeping

belak:
This allows us to drop the shutdownPollInterval without a large increase in complexity.

PR #77, opened 2018-05-07
Wait for connections to finish when shutting down

jbarnette:
PR #74 introduced a WaitGroup for listeners, but it doesn't wait for open connections before closing the server. This patch waits until all conns are closed before returning from Shutdown.

/cc @belak

### DIFF
--- a/server.go
+++ b/server.go
@@ -39,6 +39,7 @@ type Server struct {
 	mu         sync.Mutex
 	listeners  map[net.Listener]struct{}
 	conns      map[*gossh.ServerConn]struct{}
+	connWg     sync.WaitGroup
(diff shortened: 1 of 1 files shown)


belak:
Not 100% sure how to trigger a rebuild :/ I was hoping re-opening it would.

In any sense, this looks correct.

belak:
Ok, build failure is expected. Newer versions of x/crypto/ssh need math/bits which was added in go 1.9.

belak:
There's technically a race here (we should call wg.Add(1) before starting the goroutines, but that can be handled in a refactor down the road as it was a pre-existing issue). Thanks for the PR!

PR #86, opened 2018-11-02
Directly expose the SSH server KEXT, MAC and Cipher algorithms

belak:
I believe this is one of the last pieces needed for https://github.com/go-gitea/gitea/pull/3896

### DIFF
--- a/server.go
+++ b/server.go
@@ -33,6 +33,19 @@ type Server struct {
 	IdleTimeout time.Duration // connection timeout when no activity, none if empty
 	MaxTimeout  time.Duration // absolute connection timeout, none if empty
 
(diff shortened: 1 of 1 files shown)


belak:
Note that I may end up dropping this if we don't need this for gitea.

belak:
I did consider exposing the raw SSH config. I suppose that makes more sense since it should only be needed in very specific cases.

Do you have any opinions on copying the config vs just referencing it? I find it a little odd that we create a new server config for every connection.

shazow:
I don't have strong feelings either way, just an idea. :)

progrium:
I could go either way.

PR #87, opened 2018-11-12
Remote forwarding

eliastor:
Tiny fixes for #60.
BTW I don't know how to contribute to existing PR, so  if it's critical plz help me.

some other polishing and remote forwarding handler will be done a bit later


progrium:
So is the idea we close #60 for this one?

eliastor:
Yes. It's include all commits from #60.

progrium:
Shall we merge this and get it over with? The only API changes are for request handlers, which isn't really part of core usage.

eliastor:
ok) When we can see it in master?)

progrium:
Oh whoops I see what I did wrong.

PR #89, opened 2018-11-16
feat: configurable server handlers

### LINKED ISSUE #71: Alternative channel handlers
Continuing discussion #47

FYI, I'm using a modified version of gliderlabs/ssh to allow me using a custom channel handler

Here is the current diff state: https://github.com/moul/sshportal/compare/dev/moul/diff-gliderlabs-upstream?expand=1

The good points using this method are:
* we can create/use custom ChannelHandler, not depending on current interface ([example](https://github.com/moul/sshportal/tree/master/pkg/bastionsession))
* the channel handler can be decided conditionally at runtime, I'm currently using different channel handlers based on dynamic conditions ([example](https://github.com/moul/sshportal/blob/7c4aab34ed20f435be594105a6e8de47ade85abf/ssh.go#L114-L174))
* minimal API change, no compatibility break (expected)

Do you want I open a PR so we can discuss or do you prefer to talk here first?

*To be honest with you, I'm quite satisfied with the global result but not so much about my project architecture*

moul:
fixes #71

### DIFF
--- a/server.go
+++ b/server.go
@@ -210,7 +210,7 @@ func (srv *Server) SetChannelHandler(kind string, handler ChannelHandler) {
 	srv.channelHandlers[kind] = handler
 }
 
-func (srv *Server) GetChannelHandler(kind string) ChannelHandler {
+func (srv *Server) ChannelHandler(kind string) ChannelHandler {
 	srv.ensureHandlers()
 	return srv.channelHandlers[kind]
 }
--- a/server.go
+++ b/server.go
@@ -34,7 +34,7 @@ type Server struct {
 	IdleTimeout time.Duration // connection timeout when no activity, none if empty
 	MaxTimeout  time.Duration // absolute connection timeout, none if empty
 
-	channelHandlers map[string]channelHandler
+	channelHandlers map[string]ChannelHandler // fallback channel handlers
 	requestHandlers map[string]RequestHandler
 
 	listenerWg sync.WaitGroup
@@ -48,8 +48,7 @@ type RequestHandler interface {
 	HandleRequest(ctx Context, srv *Server, req *gossh.Request) (ok bool, payload []byte)
 }
 
-// internal for now
-type channelHandler func(srv *Server, conn *gossh.ServerConn, newChan gossh.NewChannel, ctx Context)
+type ChannelHandler func(srv *Server, conn *gossh.ServerConn, newChan gossh.NewChannel, ctx Context)
 
 func (srv *Server) ensureHostSigner() error {
 	if len(srv.HostSigners) == 0 {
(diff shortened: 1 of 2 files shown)


moul:
you can see it in action here: https://github.com/moul/sshportal/blob/0f294cd62d625a0ef5934d761069f97a82c9e640/main.go#L156-L173

basically, I'm saving the hold channel handlers (because they are not exported and manipulating local vars)

and then I setup my old channel handler that will in certain case do something very different than the default behavior from gliderlabs/ssh, and sometimes reuse the default handlers

server.go
```diff
@@ -65,13 +64,17 @@ func (srv *Server) ensureHostSigner() error {
 func (srv *Server) ensureHandlers() {
 	srv.mu.Lock()
 	defer srv.mu.Unlock()
-	srv.requestHandlers = map[string]RequestHandler{
-		"tcpip-forward":        forwardedTCPHandler{},
-		"cancel-tcpip-forward": forwardedTCPHandler{},
+	if srv.requestHandlers == nil {
+		srv.requestHandlers = map[string]RequestHandler{
+			"tcpip-forward":        forwardedTCPHandler{},
+			"cancel-tcpip-forward": forwardedTCPHandler{},
```
josegonzalez: Should we force set the handlers if they don't exist in the map?

josegonzalez:
Seems somewhat reasonable, though I defer to @progrium here :)

server.go
```diff
@@ -206,6 +203,18 @@ func (srv *Server) Serve(l net.Listener) error {
...
 }
 
+func (srv *Server) SetChannelHandler(kind string, handler ChannelHandler) {
+	srv.ensureHandlers()
+	srv.mu.Lock()
+	defer srv.mu.Unlock()
+	srv.channelHandlers[kind] = handler
+}
+
+func (srv *Server) GetChannelHandler(kind string) ChannelHandler {
```
progrium: It's more idiomatic to drop the "Get" in getters in Go.

progrium:
Getter/setter means you can't easily set them up when you create a Server struct, but I guess that's ok. Minor fix and I'll merge. Thanks @josegonzalez for review.

server.go
josegonzalez: I've updated the PR.

josegonzalez:
@progrium We can expose the `ChannelHandlers` map in the `Server` struct, but that would break the api. I'm also not sure what should happen if you set the `ChannelHandlers` map after a server has started - can we somehow catch that, or is that just assumed to be bad practice/won't work?

progrium:
It's fine to break API pre 1.0, just need to cut a release so people can reference the old API. 

I'm not entirely sure about that API design question though. I seem to have forgotten best idioms here. @shazow?

belak:
Configuration is a larger question since we've already got multiple different styles of configuring the server (option functions, struct values). It would be best to pick one and stick to it and just live with the trade-offs.

progrium:
If I were to pick I think I'd say struct values.

progrium:
Yeah. With so much configuration, it *might* be nice to break things down into sub-structs? Maybe a struct for all the channel handlers. One for policy callbacks (discussed a long time ago).

bgardiner:
Wanted to ping this thread to see if there is any progress here?  Looks like there is no major objection to this PR -- can we merge it?  I am building an ssh proxy server and would love to be able to use `github.com/gliderlabs/ssh` instead of the fork at `github.com/moul/ssh`.  Thanks for your work @moul

belak:
I cleaned some stuff and expanded on it in #108 - would something like that work for you?

bgardiner:
Thanks for the quick reply @belak. #108 looks great and works well for my desired use case

PR #92, opened 2019-01-06
Added Server.KeyboardInteractiveHandler

aidansteele:
I needed this in order to implement functionality similar to https://github.com/FiloSottile/whosthere#how-it-works (i.e. I wanted to print a custom error message for users who failed to present any acceptable public keys), so I added this handler.

I'm submitting it for your consideration because having it upstream means less work for _me_, but understand that it's pretty niche and might not be in the spirit of this project. Feel free to use or not :)

### DIFF
--- a/ssh.go
+++ b/ssh.go
@@ -2,7 +2,7 @@ package ssh
 
 import (
 	"crypto/subtle"
-	"golang.org/x/crypto/ssh"
+	gossh "golang.org/x/crypto/ssh"
 	"net"
 )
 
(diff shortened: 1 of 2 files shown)


ssh.go
```diff
@@ -2,6 +2,7 @@ package ssh
 
 import (
 	"crypto/subtle"
+	"golang.org/x/crypto/ssh"
```
belak: This should probably be imported as gossh.

aidansteele: Nice catch, I've fixed that now

belak:
I'm fine with merging this in for now, but note that it may be changed at a later date since we're looking at revamping how config works. Thanks for your contribution!

aidansteele:
That’s fair enough, and thank you! :100:

PR #93, opened 2019-01-24
Add way to fetch the unparsed command string from a session.

andrewchambers:


progrium:
As reasonable as this is, what do other people think about this? You can basically get this by rejoining the split command string. I'm curious, @andrewchambers what is the use case for this? I suspected the common case would be like you're using os.Args and nobody has requested something like this until now.

andrewchambers:
I am forwarding to another ssh server, and letting it handle things like:

```FOO=bar command``` 

I didn't trust shlex to preserve all bash semantics, even though perhaps it might, I couldn't guarantee it, so did this to ensure it works in every case.

belak:
At my old job I would have been looking for this as well - there was some custom parsing we wanted to do with the commands that wasn't possible when it came through shlex

belak:
This was merged as a part of #110. Thanks for submitting this!

PR #95, opened 2019-02-21
Add DefaultServerConfigCallback option for create custom default `ServerConfig`s

binwiederhier:
This callback allows a more granular control of the SSH server configuration, such as for the key exchange / cipher / MAC algorithms.

Example usage:

```
server := ssh.Server{
  Addr: ":2222",
  // ...
  DefaultServerConfigCallback: func (ctx ssh.Context) *gossh.ServerConfig {
    config := &gossh.ServerConfig{}

    config.KeyExchanges = []string{"curve25519-sha256@libssh.org"}
    config.MACs = []string{"hmac-sha2-256-etm@openssh.com"}
    config.Ciphers = []string{
      "aes128-gcm@openssh.com",
      "ssh-ed25519",
      "ssh-rsa",
    }

    return config
  },
}

server.ListenAndServe()
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -30,7 +30,8 @@ type Server struct {
 	PtyCallback                   PtyCallback                   // callback for allowing PTY sessions, allows all if nil
 	ConnCallback                  ConnCallback                  // optional callback for wrapping net.Conn before handling
 	LocalPortForwardingCallback   LocalPortForwardingCallback   // callback for allowing local port forwarding, denies all if nil
-	ReversePortForwardingCallback ReversePortForwardingCallback //callback for allowing reverse port forwarding, denies all if nil
+	ReversePortForwardingCallback ReversePortForwardingCallback // callback for allowing reverse port forwarding, denies all if nil
+	DefaultServerConfigCallback   DefaultServerConfigCallback   // callback for configuring detailed SSH options
 
 	IdleTimeout time.Duration // connection timeout when no activity, none if empty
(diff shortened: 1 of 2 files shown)


progrium:
LGTM

belak:
Huh, interesting. I like this approach. This would also replace https://github.com/gliderlabs/ssh/pull/86

belak:
Alternatively, this could be done by adding a config hook which would take a context and the ServerConfig and allow the user to modify it before actually using it.

binwiederhier:
I didn't see that #86 existed. Whatever you guys think is best. :-)

progrium:
I like callbacks. Eventually will clean up the API into single delegate that can implement various interfaces. We/somebody can change this to pass in a default config later.

binwiederhier:
Awesome. You guys are super quick. How long does it usually take to get things merged / into a tag?

progrium:
I'd have merged already if @belak explicitly gave thumbs up. Tagging is done by request right now.

belak:
As long as we understand that the API may change later, I'm fine with this in its current form.

progrium:
Pre-1.0 the API will be changing a lot

belak:
Then I'm good with this. :+1: from me!

binwiederhier:
Awesome thanks guys. Thank was so quick! 
Can I get a tag too? :heart:

belak:
Tagged as v0.1.3. Thanks for contributing!

Issue #99, opened 2019-03-31
bug: connection killed whenever MaxTimeout is not set but IdleTimeout is


czoczo:
This is similar to #51 but other way around.
### Way to reproduce:
```
set value for IdleTimeout, e.g. 10 sec
omit value for MaxTimeout
```
### What is expected:
When connection is idle for 10 sec, it should be killed
### What actually happens:
Connection gets killed after 10 sec no matter how active the connection is

belak:
Fixed in 9253a53eda6d496306693fc247b7819a204040c1

PR #100, opened 2019-03-31
Fix connection kill, when IdleTimeout given without setting MaxTimeout.

czoczo:
[This](https://github.com/gliderlabs/ssh/pull/100/commits/cc3893658065ff6da1ebd73c4ed2aa659aa5b020) solves #99 
### explanation:
Whenever IdleTimeout is set and MaxTimouet is not, [serverConn](https://github.com/gliderlabs/ssh/blob/ef6d89046be36104109e42ac1fee6601f9be95d7/conn.go#L9) struct initiates maxDedline with `time.Time` type.
conn.go:
```
type serverConn struct {
    net.Conn

    idleTimeout   time.Duration
    maxDeadline   time.Time
    closeCanceler context.CancelFunc
}
```
`time.Time` is set to zero value by default, thus making the if in [updateDeadline](https://github.com/gliderlabs/ssh/blob/ef6d89046be36104109e42ac1fee6601f9be95d7/conn.go#L43) function never fire off:
```
/* c.maxDeadline is zero */
if idleDeadline.Unix() < c.maxDeadline.Unix() {
            c.Conn.SetDeadline(idleDeadline)
            return
        }
    (...)
```
[My pull request](https://github.com/gliderlabs/ssh/pull/100/commits/cc3893658065ff6da1ebd73c4ed2aa659aa5b020) fixes that bug. Pleas consider merging my change.

P.S. That's my first pull request ever. Please comment if anything is not understandable or not up to your standards.


belak:
Thanks for your contribution! Good catch!

This was fixed by cherry-picking in 9253a53eda6d496306693fc247b7819a204040c1.

In the future, it makes our lives much easier if you use a separate branch (other than master) for your PR and don't update the README to remove all the text. If you use a separate branch, you could update the README on master so it shows up on your fork without also submitting those changes as a part of your pull request.

czoczo:
Wow - I'm impressed by your reaction time! Sorry for any inconveniences. Many thanks for the hints - all duly noted :)

PR #104, opened 2019-04-27
wait for child process after client disconnect

kuehnelth:
Without Wait() the child process stays as a zombie process until the server process terminates.


belak:
Good catch! Thanks!

PR #108, opened 2019-06-12
Configurable handlers

belak:
This cleans up the configurable-handlers branch, brings some of our handler code more in line with what the stdlib does, and disables port forwarding by default with an option to enable it (we *really* shouldn't have shipped it like this by default).

This does break backwards compatibility (handler changes specifically) but adds quite a bit of flexibility without sacrificing complexity for people who just want to use this package out of the box.

This directly replaces #89

### DIFF
--- a/server.go
+++ b/server.go
@@ -37,8 +37,15 @@ type Server struct {
 	IdleTimeout time.Duration // connection timeout when no activity, none if empty
 	MaxTimeout  time.Duration // absolute connection timeout, none if empty
 
-	channelHandlers map[string]ChannelHandler // fallback channel handlers
-	requestHandlers map[string]RequestHandler
+	// ChannelHandlers allow overriding the built-in session handlers or provide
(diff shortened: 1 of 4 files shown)


bgardiner:
Thanks @belak. This looks great to me. I like exposing the `ChannelHandlers` and `RequestHandlers` maps publicly on the `Server` struct as opposed to the getters and setters in #89

gustavosbarreto:
Waiting for this pull request to be merged

belak:
Hey all, I don't really have any projects to test this with any more - though that will be changing when I find the time. If you get a chance to check it and it works for you, let me know and I'll see about merging this.

belak:
It's been a while and we've got people using similar setups already, so I feel comfortable merging this. Next tag will be a minor version bump (as with semver 0.x I'm pretty sure minor bumps can be breaking) rather than a patch version.

bgardiner:
This is working great for me so far in a proxy server I wrote.  Thanks for developing this and merging @belak

PR #109, opened 2019-06-17
Fix forwardedTCPHandler

gustavosbarreto:
This fixes an issue when a forwarding client disconnects from the server, the listening port remains open because `forwards map[string]net.Listener` are never changed due a mutability issue in `forwardedTCPHandler`


belak:
Pushed in 86efb7062814fb833c9105be36a6a7f213a935ad

Note that it's not enabled by default, so only half of this change was needed. Thanks for your contribution!

PR #110, opened 2019-06-19
Add Session.RawCommand()

belak:
My main use case for this would be SSH_ORIGINAL_COMMAND. Many SSH servers set this when running commands on the server itself, and it's very annoying (and error prone) to set without having the RawCommand. I also opted for only calling shlex when needed, as it's an operation which doesn't need to be done in all cases.

This would replace #93 (Proper attribution for @andrewchambers is still there) as the original pull request is missing the line to actually set rawCmd.

### DIFF
--- a/session.go
+++ b/session.go
@@ -44,6 +44,9 @@ type Session interface {
 	// which considers quoting not just whitespace.
 	Command() []string
 
+	// RawCommand returns the exact command that was provided by the user.
+	RawCommand() string
+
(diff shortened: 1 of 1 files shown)


belak:
I do sort of wish it was possible to make shlex optional if it's not needed, but this at least removes the call.

andrewchambers:
Thanks for doing that, I sort of just used my own branch and forgot, cheers :)

PR #111, opened 2019-06-19
Add subsystem support

belak:
This subsystem support is based on the new handler work, but allows for flexible subsystem handlers at the server level.


charl:
Any chance of this moving ahead to add subsystem support?

belak:
Sure! I don't see any reason why not. However, this is a very early implementation - if you run into issues, please let me know and I'll do my best to fix them.

PR #120, opened 2019-09-27
feat: Make the HandleConn method public.

chirino:
This will let you do more whacky things like running ssh over tls, etc.


progrium:
hell yea

joshwget:
Great change 👍

PR #122, opened 2019-10-08
Passes Context into ConnCallback

Michael9127:
This moves context creation before the ConnCallback invocation and passes it into the callback. Allowing state to be set for the duration of the connection.

### DIFF
--- a/options_test.go
+++ b/options_test.go
@@ -95,7 +95,7 @@ func TestConnWrapping(t *testing.T) {
 		HostKeyCallback: gossh.InsecureIgnoreHostKey(),
 	}, PasswordAuth(func(ctx Context, password string) bool {
 		return true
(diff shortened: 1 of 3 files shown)


belak:
I believe by convention the context is passed in as the first argument, otherwise, looks good to go!

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

Michael9127:
fixed, thanks!

belak:
Awesome! Thanks! This is a breaking change so the first tagged version to have this will probably be 0.3.

rtucek:
@belak would you mind tagging this as 0.3?
I'm eagerly waiting for this hook for injecting ctx values as soon as a client connects.

belak:
Yep, done.

PR #123, opened 2019-10-16
Update AddHostKey to avoid always appending

belak:
As it mentions in the comment, this is based on the code from x/crypto/ssh and is done to optimize dynamically loading keys at runtime (to avoid keeping old keys around). I've been experimenting and I would have just done it with the config callback, but it seems like there's not a great way to do this (since we auto-generate a key if there are none, and always add the HostSigners to the server config).

The config at runtime has a number of race conditions still present (they mostly happen if you set an option at runtime directly) but the one for AddHostKey should be fixed.

### DIFF
--- a/server.go
+++ b/server.go
@@ -315,6 +315,18 @@ func (srv *Server) ListenAndServe() error {
 func (srv *Server) AddHostKey(key Signer) {
 	// these are later added via AddHostKey on ServerConfig, which performs the
 	// check for one of every algorithm.
+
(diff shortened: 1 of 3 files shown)


belak:
There seems to be a test that's hanging... so I'll have to look into that.

belak:
Ok, I rewrote TestSignals - I couldn't get t.Fatal to work from inside a goroutine, so I passed errors out.

PR #125, opened 2019-10-22
Server able to close all connections for given user

nerd2:


belak:
I really like the idea behind this... can you think of any way to make this more generic? As an example, I've been experimenting with a git server in go, but it would be hard to use this there because traditionally everyone SSHs in as the `git` user even though we tag them later.

server.go
```diff
@@ -323,6 +326,19 @@ func (srv *Server) SetOption(option Option) error {
 	return option(srv)
 }
 
+func (srv *Server) CloseConnectionsForUser(user string) {
```
kissgyorgy: This is too specific to your use-case. A better way to generalize this would be to make connections exported or a hook to get all connections or something, so the specific implementations can do whatever they want with all the connections.

PR #130, opened 2019-11-15
Add localforward example

XSAM:
I want to implement an SSH server to forward a remote server's port to local (aka *local forward*), then I found an example called **remoteforward**, and it uses `LocalPortForwardingCallback` setting.

When I try to use the example to bind local port and use `curl` to request,

```bash
ssh -L 0.0.0.0:8080:example.com:80 localhost -p 2222
curl localhost:8080
```

Got error

```
channel 2: open failed: unknown channel type: unsupported channel type
```

Then I found it cannot be used in *local forward*, and the `LocalPortForwardingCallback` in this example kinda useless and bring some confuse.

So I remove the `LocalPortForwardingCallback` in **remoteforward** example and add a new example called **localforward**


RangelReale:
Thanks, the single existing forward example was confusing.

But this still don't work for me, using `ssh -L 0.0.0.0:8080:example.com:80 localhost -p 2222` I'm still getting to the `session` handler, instead of the local forwarding.

PR #132, opened 2019-12-24
go.mod(s)

wxio:
Created mulitple go.mod(s) for the library and examples.
Only for the examples which have more requirements than the library.
This way clients don't get extra dependances.


belak:
Thanks for submitting this. go.mod support landed in #153.

PR #140, opened 2020-09-24
DirectTCPIPHandler func use payload origin addr and port

eudore:
The `DirectTCPIPHandler` function establishes a new connection using the `net.TCPAddr` function so that the address and port of Origin can be set.

```golang
func DirectTCPIPHandler(srv *Server, conn *gossh.ServerConn, newChan gossh.NewChannel, ctx Context) {
	d := localForwardChannelData{}
	...
	laddr := &net.TCPAddr{
		IP:   net.ParseIP(d.OriginAddr),
		Port: int(d.OriginPort),
	}
	raddr := &net.TCPAddr{
		IP:   net.ParseIP(d.DestAddr),
		Port: int(d.DestPort),
	}
	dconn, err := net.DialTCP("tcp", laddr, raddr)
	...
}
```

### DIFF
--- a/tcpip.go
+++ b/tcpip.go
@@ -37,10 +37,15 @@ func DirectTCPIPHandler(srv *Server, conn *gossh.ServerConn, newChan gossh.NewCh
 		return
 	}
 
-	dest := net.JoinHostPort(d.DestAddr, strconv.FormatInt(int64(d.DestPort), 10))
-
-	var dialer net.Dialer
-	dconn, err := dialer.DialContext(ctx, "tcp", dest)
+	laddr := &net.TCPAddr{
+		IP:   net.ParseIP(d.OriginAddr),
+		Port: int(d.OriginPort),
+	}
+	raddr := &net.TCPAddr{
+		IP:   net.ParseIP(d.DestAddr),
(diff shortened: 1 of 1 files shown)


tcpip.go
```diff
@@ -37,10 +37,15 @@ func DirectTCPIPHandler(srv *Server, conn *gossh.ServerConn, newChan gossh.NewCh
...
-	dconn, err := dialer.DialContext(ctx, "tcp", dest)
+	laddr := &net.TCPAddr{
+		IP:   net.ParseIP(d.OriginAddr),
+		Port: int(d.OriginPort),
+	}
+	raddr := &net.TCPAddr{
+		IP:   net.ParseIP(d.DestAddr),
+		Port: int(d.DestPort),
+	}
+	dconn, err := net.DialTCP("tcp", laddr, raddr)
```
belak, line 48: This seems like it would be a worthwhile change (though I'm not 100% sure I understand it), but I'd like for it to be done without dropping ctx from the dial call.

I believe this can be done by setting `dialer. LocalAddr` to the `laddr` and using the previous method of creating the `raddr`.

PR #141, opened 2020-10-05
Register chan to Session to listen for break requests

meislerj:
This change supports break over SSH (https://tools.ietf.org/html/rfc4335) by adding a method to Session to register a channel to indicate that a break request was received.

### DIFF
--- a/session.go
+++ b/session.go
@@ -359,11 +359,13 @@ func (sess *session) handleRequests(reqs <-chan *gossh.Request) {
 			req.Reply(true, nil)
 		case "break":
 			ok := false
+			sess.Lock()
 			if sess.breakCh != nil {
 				sess.breakCh <- true
 				ok = true
 			}
 			req.Reply(ok, nil)
+			sess.Unlock()
 		default:
 			// TODO: debug log
 			req.Reply(false, nil)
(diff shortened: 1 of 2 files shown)


meislerj:
Hi. I'm working on a project that uses gliderlabs/ssh and needs to handle break requests. I've implemented this PR to add break support in a way that feels consistent with the existing Signals design. Thanks in advance for your help!

belak:
This is interesting. I wasn't aware of BREAK.

I think this is missing session.Lock() and sess.Unlock() around checking the value to session.breakCh, so there's a possible race condition.

meislerj:
Good call! I added it in the second commit.

meislerj:
Just wanted to check on this. Do you have other questions or concerns? Thanks!

belak:
No, this looks good to me - thanks for submitting this!

meislerj:
Thanks for the help!

meislerj:
Would you mind creating a new tag for this? I see the last tag is from February. Thanks!

belak:
> Would you mind creating a new tag for this? I see the last tag is from February. Thanks!

Done, I've tagged master as v0.3.1

PR #146, opened 2021-02-01
Fix: KeyboardInteractive Login

Lemonn:
To be able to use the KeyboardInteractive method a few thinks need to be added

+ Include KeyboardInteractive in decision if NoClientAuth should be performed
+ Add option for KeyboardInteractive


belak:
Looks good to me. Thanks!

belak:
I've tagged this as v0.3.2

PR #148, opened 2021-03-28
Prevent Context.RemoteAddr() from exploding when called from ConnCallback

amery:
When calling ctx.RemoteAddr() within ConnCallback one gets a panic due to
unsafe casting within the accessor. I understand it's not a valid scenario
as such, but I accidentally called a logger that expected that checked
for RemoteAddr() in ssh.Context

panic: interface conversion: interface is nil, not net.Addr

Signed-off-by: Alejandro Mery <amery@geeks.cl>

### DIFF
--- a/context.go
+++ b/context.go
@@ -140,7 +140,10 @@ func (ctx *sshContext) ServerVersion() string {
 }
 
 func (ctx *sshContext) RemoteAddr() net.Addr {
-	return ctx.Value(ContextKeyRemoteAddr).(net.Addr)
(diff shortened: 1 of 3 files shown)


amery:
uhm...

```
#!/bin/bash -eo pipefail
go get
go: cannot find main module, but found .git/config in /go/src/github.com/gliderlabs/ssh
	to create a module there, run:
	go mod init

Exited with code exit status 1
CircleCI received exit code 1
```

belak:
Thanks for submitting this! I've cherry picked your commit and it's now in master as c7b4a1b.

PR #153, opened 2021-06-17
Add go mod files

rumpelsepp:
An updated version of #132.


belak:
Thanks for submitting this! In hindsight we should have gotten to this quite a bit sooner.

PR #154, opened 2021-08-17
readme corregido

ivansoriasolis:


gustavosbarreto:
As a native Portuguese speaker, I perfectly understand your commit message even written in Spanish but commit messages should be always in English.

PR #161, opened 2021-10-03
fix: retain auth config from ServerConfigCallback

Mohammed90:
The previous code overrides the authentication callbacks of ServerConfigCallback with the server authentication config, which loses the customized config per-connection. This connection checks for the presence of per-connection authentication config and only overrides it if it's nil.

Moreover (and we can move the discussion to the following to an issue), what do you think about swapping the priority between config from `ServerConfigCallback` and the server config? Basically, instead of the server config overriding the config from `ServerConfigCallback`, then the config from `ServerConfigCallback` overrides the server. The reasoning is the `ServerConfigCallback` is the only place where the user of the package can customize the config per connection/client basis. Without the proposed swap, it's impossible to emulate the `Match` block in [sshd_config(5)](https://man7.org/linux/man-pages/man5/ssh_config.5.html).

This is perhaps a breaking-change, so it's definitely subject for discussion.

### DIFF
--- a/server.go
+++ b/server.go
@@ -132,7 +132,7 @@ func (srv *Server) config(ctx Context) *gossh.ServerConfig {
 	if srv.Version != "" {
 		config.ServerVersion = "SSH-2.0-" + srv.Version
 	}
-	if srv.PasswordHandler != nil {
+	if srv.PasswordHandler != nil && config.PasswordCallback == nil {
 		config.PasswordCallback = func(conn gossh.ConnMetadata, password []byte) (*gossh.Permissions, error) {
 			applyConnMetadata(ctx, conn)
 			if ok := srv.PasswordHandler(ctx, string(password)); !ok {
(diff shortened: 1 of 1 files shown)


belak:
Interesting tweak - I'd be interested to hear your use case for setting an auth Callback on both the server and gossh.ServerConfig. Would it be possible to change your code to only set auth callbacks by using the ServerConfigCallback? I think the original idea was that if you return auth callbacks from the ServerConfigCallback and don't set `srv.*Callback`, we wouldn't override them.

Maybe changing this to also check `config.*Callback` would be a better fix for accidentally setting NoClientAuth.
```
	if srv.PasswordHandler == nil && srv.PublicKeyHandler == nil && srv.KeyboardInteractiveHandler == nil {
		config.NoClientAuth = true
	}
```

Additionally, if we are supporting user-provided `config.*Callback` functions moving forward, it would probably make sense to wrap them, so they also set the conn metadata (using `applyConnMetadata`).

Also, if I'm reading this right, we never actually set the Permissions so Context can use them.

This is not very fun, stringy code, so I'm not actually 100% sure what the best route forward is, but because this library has set a precedent (overriding the gossh.ServerConfig with `srv.*Callbacks`) I'm hesitant to change the pattern.

One other thing worth noting - the *whole* API for gliderlabs/ssh could use some reworking from what has been learned since this library was released. I've considered forking/rewriting (and putting it under https://github.com/go-ssh/ssh) but I haven't had the time yet.

Sorry, that was a bit more of a rambling response than I originally intended, but it's been a while since I've looked at this library.

mohammed90:
Following our conversation over email, it's best to retain the current behavior. Clients having similar needs to mine should own the provision of default config as well as customized config via `ServerConfigCallback`.

It's worth noting providing custom config based on the current implementation because the `ssh.Context` isn't filled with data until:

https://github.com/gliderlabs/ssh/blob/c9fc4412c4220e738f2c504b07b7c7dcc498108d/server.go#L135-L162

while the call to `ServerConfigCallback` is done on:
https://github.com/gliderlabs/ssh/blob/c9fc4412c4220e738f2c504b07b7c7dcc498108d/server.go#L124

Moreover, the `gossh.ConnMetadata` is not available until the authentication callbacks, not before, which means best effort of amendments to this library will only allow for decisions based on `RemoteAddr() net.Addr` and `LocalAddr() net.Addr` because they are available on `net.Conn` which is received from the listener

https://github.com/gliderlabs/ssh/blob/c9fc4412c4220e738f2c504b07b7c7dcc498108d/server.go#L262-L281

Other metadata, e.g. the user, is only available once x/crypto/ssh parses the bytes from the connection which is only done within/after the call to `gossh.NewServerConn`, by then it's too late.

PR #163, opened 2021-11-07
Consider result of `ServerConfigCallback` when setting NoClientAuth

mohammed90:
I ran into this when I was controlling the authentication methods on per connection basis. Effectively `NoClientAuth` ends up being `true` for all connections, regardless of the result `ServerConfigCallback`, because the authentication callbacks on the server are not configured for the lack of global authentication approach. This commit only disables client authentication when it's explicitly disabled or the authentication callkbacks on both `*Server` and `ServerConfig` are `nil`.

### DIFF
--- a/server.go
+++ b/server.go
@@ -126,9 +126,9 @@ func (srv *Server) config(ctx Context) *gossh.ServerConfig {
 	for _, signer := range srv.HostSigners {
 		config.AddHostKey(signer)
 	}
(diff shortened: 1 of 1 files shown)


belak:
This seems like a good solution. I'll try to take a look and get this merged in the next few days.

gustavosbarreto:
@belak PING

Issue #165, opened 2021-11-16
Support RSA SHA-2 (RFC8332) signatures


SURA907:
Cf golang/go#37278

Crypto already supports RSA SHA-2 (RFC8332) signatures

Only need to upgrade mod to support RSA SHA-2 (RFC8332) signatures

Thanks

SURA907:
It looks like x/crypto/ssh is not ready

mikesmitty:
Looks like x/crypto/ssh support added support recently. Could this be added now?
https://github.com/golang/crypto/commit/5d542ad81a58c89581d596f49d0ba5d435481bcf

belak:
Is it really just updating the mod that's needed? That should be an easy fix.

quackduck:
We do not have server support yet, so we're still waiting for that change to happen in x/crypto

Client support is already here and simply requires a `go get -u golang.org/x/crypto`

quackduck:
Switching from x/crypto to [rmohr/crypto](https://github.com/rmohr/crypto) would add both client and server support and fix this issue. The problem I see is that we would then be using a fork not officially maintained by the Go project and would be a bit out of date.

gustavosbarreto:
Until `gliderlabs/ssh` package updates its direct dependency on `crypto/ssh` you have to update its on your `go.mod` project file by running `go get -u golang.org/x/crypto` in the project root dir.

belak:
I *think* this should be fixed now - golang.org/x/crypto was updated and a new version (v0.3.5) has been tagged.

erwin:
Perhaps I'm doing something wrong, but I've still been unable to connect when using an RSA key (rsa-sha2-256, rsa-sha2-512) unless I turn on the deprecated `-oPubkeyAcceptedKeyTypes=+ssh-rsa` flag...

#### Step 1: Build the ssh-publickey example, with latest versions of gliderlabs/ssh and x/crypto

```
cd /tmp/test
git clone https://github.com/gliderlabs/ssh/
cd ssh/_examples/ssh-publickey
go mod init github.demo/_examples/ssh-publickey
go get github.com/gliderlabs/ssh@latest
go get golang.org/x/crypto@latest
go build .
./ssh-publickey
```

#### Step 2: Then test connecting with an RSA key:

```
cd /tmp/test
ssh-keygen -q -t rsa -N "" -f test-key-rsa
[...]

PR #166, opened 2021-11-16
Support RSA SHA-2 (RFC8332) signatures(#165)

SURA907:
* upgrade golang.org/x/crypto


SURA907:
It looks like x/crypto/ssh is not ready,  this pr will be closed
I will continue to follow `golang.org/x/crypto` and open pr again when it is ready
cf: https://github.com/golang/go/issues/37278
cf: https://github.com/golang/crypto/pull/197

belak:
Thanks for the update. I'll also follow that PR/issue for when it's ready.

PR #167, opened 2021-11-25
Allow client to use port 0 when requesting reverse port forwarding

rawoul:
Bind the port to forward before calling the ReversePortForwardingCallback
callback, with the actual bound port instead of 0.


felipereyel (approved):

felipereyel:
having the same issue here

PR #168, opened 2022-01-06
feat: return ssh.Context

caarlos0:
changes `ssh.Session` interface to return `ssh.Context` instead of `context.Context`.

This allows users to do things like `session.Context().SetValue(foo, bar)`.


belak:
This should be alright since it's just widening the return type and should work everywhere context.Context already does. Thanks!

PR #103, opened 2019-04-26
Support subsystem

ibuler:
Hi, 

Subsystem support added, see examples

#63 
#101

PR #182, opened 2022-10-01
add sftp-server example

jm33-m0:
Since subsystem support was added, I think it will be helpful to add an SFTP server example as it's the most common use case for subsystem and not so straight-forward to implement.

### DIFF
--- a/_examples/ssh-sftpserver/sftp.go
+++ b/_examples/ssh-sftpserver/sftp.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+	"fmt"
+	"io"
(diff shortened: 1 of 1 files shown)


jm33-m0:
Code copied from #103, and adapted for #111.
image

gustavosbarreto:
I am trying to figure out why the CI is failing but no luck yet.

@belak any idea?

jm33-m0:
@gustavosbarreto The code builds fine, as shown in the screenshot. Since this is just an example addition, I wouldn't worry about the CI results

PR #185, opened 2022-10-21
feat: replace newlines with crlf newlines when in a pty

caarlos0:
Writing to the stdout of the session already handles replacing `\n` with `\r\n`, but the stderr didn't had the same treatment.

Created another `io.Writer` wrapper that can be used in both, and a "session.SafeStderr()` that returns it.

PS: pretty sure "safe writer" is not a good name, also considered "crlf writer", let me know if you like one better than the other, or maybe yet another name


session.go
```diff
@@ -82,6 +82,12 @@ type Session interface {
 	// the request handling loop. Registering nil will unregister the channel.
 	// During the time that no channel is registered, breaks are ignored.
 	Break(c chan<- bool)
+
+	// Stderr returns an io.ReadWriter that writes to this channel
+	// with the extended data type set to stderr. Stderr may
+	// safely be read and written from a different goroutine than
+	// Read and Write respectively.
+	Stderr() io.ReadWriter
```
belak, line 90: Shouldn't this be a plain io.Writer? I'm fairly certain you can't read data from stderr.

caarlos0: IMHO it could, but it comes as a readwriter from ssh.Channel: https://pkg.go.dev/golang.org/x/crypto/ssh#Channel

which we embed on our session here: https://github.com/gliderlabs/ssh/blob/db094653c80809f63e06c08eda6987780c3c4a08/session.go#L23

That said, if I try to change it to writer only, it won't compile... so unfortunately I think there's no other way (besides maybe a new stderr method with another name).

pty.go
```diff
@@ -0,0 +1,57 @@
+package ssh
+
+import (
+	"bytes"
+	"io"
+)
+
+// NewPtyWriter creates a writer that handles when the session has a active
+// PTY, replacing the \n with \r\n.
+func NewPtyWriter(w io.Writer) io.Writer {
```
aymanbagabas, line 10: Do we need to export `NewPtyWriter` and `NewPtyReadWriter`?

aymanbagabas (approved):

pty.go
caarlos0: good question... I don't think we **need** to export, but I think it would be nicer if we do... whatever @belak decides is fine by me though.

aymanbagabas:
@belak Could you take a look at this?

PR #186, opened 2022-10-21
Fix normalize session lf newlines to crlf on stderr

aymanbagabas:
Wrap Stderr to normalize the output depending on if the session has a pty.


aymanbagabas:
closing in favor of https://github.com/gliderlabs/ssh/pull/186

PR #187, opened 2022-10-25
add PasswordAuthE func, which is like PasswordAuth but returns error instead of bool

umialpha:
Add `PasswordAuthE` function, because sometimes ssh server should returns detailed error than "permission denied".
if `PasswordAuthE`and `PasswordAuth` both set, `PasswordAuthE` will overwrite PasswordAuth

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -1,6 +1,7 @@
 package ssh_test
 
 import (
+	"fmt"
 	"io"
 	"io/ioutil"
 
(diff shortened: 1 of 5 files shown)


gustavosbarreto:
@umialpha What do you think about this callback being generic for all of authentication errors (password, publickey...)?

umialpha:
> @umialpha What do you think about this callback being generic for all of authentication errors (password, publickey...)?

ok, let me rethink how to unify them

PR #188, opened 2022-10-28
feat: use a common Err for permission denied

guilhem:
This error can be used in ConnectionFailedCallback to know error with error.Is().

### DIFF
--- a/server.go
+++ b/server.go
@@ -3,7 +3,6 @@ package ssh
 import (
 	"context"
 	"errors"
-	"fmt"
 	"net"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -29,6 +28,8 @@ var DefaultChannelHandlers = map[string]ChannelHandler{
 	"session": DefaultSessionHandler,
 }
 
+var ErrPermissionDenied = errors.New("permission denied")
```
gustavosbarreto, line 31: This could be in a single var statement with all error types.

gustavosbarreto:
Could you please write a test case for these error type?

PR #200, opened 2023-03-15
fix: support multiple command from ssh client

whatwewant:
# Example

```bash
$ ssh zero@1.2.3.4 "
ls -al
whoami
date
"
```

### DIFF
--- a/session.go
+++ b/session.go
@@ -5,9 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"strings"
 	"sync"
 
-	"github.com/anmitsu/go-shlex"
 	gossh "golang.org/x/crypto/ssh"
 )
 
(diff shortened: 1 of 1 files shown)


gustavosbarreto:
Hi @whatwewant,

I understand that for your example this makes sense, but the use of shlex.Split is necessary for much more complex use cases than simply splitting a string into slices. If there is an issue with the shlex.Split function, then it should be addressed directly in the upstream code and the dependency in go.mod updated accordingly, rather than finding a workaround in the code that may not be suitable for all use cases.

gustavosbarreto:
Hey @belak, I don't think it makes sense to simply split the string into parts without taking into account the POSIX rules of the `shlex.Split` function. If you also agree, I think this PR should be closed.

belak:
I've done a little investigating with what the expected behavior here should be - arguably this library should return a plain string and let the user parse it.

What you're seeing with the multi-line command running multiple commands (how this command reacts when connecting to an OpenSSH server) is because it's getting passed to a shell which is interpreting the string passed in as a small shell script (hence why the first line is ignored).

Arguably both how this library currently handles it, and the proposed change aren't the "correct" way of handling this form of arguments. It would take a breaking API change to change this library to do the right thing, and I don't think we want to do that right now.

Issue #160, opened 2021-08-31
Data race on sshContext


prologic:
came across this data race:

```#!sh
==================
WARNING: DATA RACE
Write at 0x00c0000988d0 by goroutine 14:
  github.com/gliderlabs/ssh.(*sshContext).SetValue()
      /Users/prologic/go/pkg/mod/github.com/gliderlabs/ssh@v0.3.3/context.go:123 +0xa6
  github.com/gliderlabs/ssh.(*Server).config.func2()
      /Users/prologic/go/pkg/mod/github.com/gliderlabs/ssh@v0.3.3/server.go:150 +0x18b
  golang.org/x/crypto/ssh.(*connection).serverAuthenticate()
      /Users/prologic/go/pkg/mod/golang.org/x/crypto@v0.0.0-20210616213533-5ff15b29337e/ssh/server.go:523 +0x2799
  golang.org/x/crypto/ssh.(*connection).serverHandshake()
      /Users/prologic/go/pkg/mod/golang.org/x/crypto@v0.0.0-20210616213533-5ff15b29337e/ssh/server.go:277 +0xb78
  golang.org/x/crypto/ssh.NewServerConn()
      /Users/prologic/go/pkg/mod/golang.org/x/crypto@v0.0.0-20210616213533-5ff15b29337e/ssh/server.go:206 +0x2ab
  github.com/gliderlabs/ssh.(*Server).HandleConn()
      /Users/prologic/go/pkg/mod/github.com/gliderlabs/ssh@v0.3.3/server.go:281 +0x2af

Previous read at 0x00c0000988d0 by goroutine 39:
  github.com/gliderlabs/ssh.(*sshContext).Done()
      <autogenerated>:1 +0x48
  net/http.http2awaitRequestCancel()
      /usr/local/Cellar/go/1.16.6/libexec/src/net/http/h2_bundle.go:6823 +0x117
  net/http.(*http2clientStream).awaitRequestCancel()
      /usr/local/Cellar/go/1.16.6/libexec/src/net/http/h2_bundle.go:6846 +0x64
```

belak:
Do you have sample code, or was this when running tests? Unfortunately, unless we happen to run across the same race, there's no way to debug this.

Also, I just noticed that this is coming from inside an http handler, so there's a larger likelihood that it's an issue with your code somewhere.

belak:
[...]

PR #209, opened 2023-07-12
fix: context data race

aymanbagabas:
Context may involve in a RW data race when setting and accessing a value and its done channel.

Fixes: https://github.com/gliderlabs/ssh/issues/160


gustavosbarreto (approved):

gustavosbarreto:
Closing this PR in favor of #211.

PR #210, opened 2023-07-31
add support for terminal opcodes

### LINKED ISSUE #98: Terminal modes
There's a line here that mentions terminal modes: https://github.com/gliderlabs/ssh/blob/e5ece1489c20cb346a08e070f7f65a3e90f83e98/ssh.go#L77

What do we envision for "terminal modes"?

aymanbagabas:
All credit goes to @maisem and Tailscale https://github.com/tailscale/tailscale/pull/4154

Fixes: https://github.com/gliderlabs/ssh/issues/98
Fixes: https://github.com/gliderlabs/ssh/pull/131

### DIFF
--- a/session.go
+++ b/session.go
@@ -346,7 +346,7 @@ func (sess *session) handleRequests(reqs <-chan *gossh.Request) {
 				req.Reply(false, nil)
 				continue
 			}
-			win, ok := parseWinchRequest(req.Payload)
+			win, _, ok := parseWindow(req.Payload)
 			if ok {
 				sess.pty.Window = win
 				sess.winch <- win
--- a/session_test.go
+++ b/session_test.go
@@ -230,9 +230,9 @@ func TestPty(t *testing.T) {
 
 func TestPtyResize(t *testing.T) {
 	t.Parallel()
-	winch0 := Window{40, 80}
-	winch1 := Window{80, 160}
-	winch2 := Window{20, 40}
+	winch0 := Window{40, 80, 320, 640}
+	winch1 := Window{80, 160, 640, 1280}
+	winch2 := Window{20, 40, 160, 320}
 	winches := make(chan Window)
 	done := make(chan bool)
 	session, _, cleanup := newTestSession(t, &Server{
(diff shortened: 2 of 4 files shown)


belak:
I want to make sure I understand this right before merging it in. @bradfitz might also be a good person to weigh in.

Additionally, could we use this to replace the hard-coded \n to \r\n translation, using something like [this block](https://github.com/tailscale/tailscale/pull/4154/files#diff-5f1a2a45f6e2d0d5d45e93930bf6a3705761f390cc95e26c99aecce59c8f9236R294-R317) in the referenced PR?

aymanbagabas:
The emulation emulates [Termios ONLCR](https://man7.org/linux/man-pages/man3/termios.3.html#:~:text=uppercase%20on%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20output.-,ONLCR,-(XSI)%20Map%20NL) mode which maps NL-CRNL (terminals 🙄).

This would only matter if the modes are used with PTY allocation. In that case, the emulation might corrupt the terminal and we would have to disable it to make the allocated PTY work correctly. A possible solution is to be able to disable the emulation. Something like https://github.com/gliderlabs/ssh/commit/4f6e0b53ea32a4030adcd068d1f680d65077639e. Otherwise, and if breaking the API is an option, users can opt-in using a `PTYWriter` interface https://github.com/gliderlabs/ssh/pull/185.

aymanbagabas:
So more down the rabbit hole of SSH/PTY support. OpenSSH client sets the terminal into _raw mode_ when the client requests a PTY session. [^1][^2] This sets the [`OPOST`](https://www.gnu.org/software/libc/manual/html_node/Output-Modes.html) output flag which enables processing the output data using flags like `ONLCR`.

This means if you have this flag disabled before establishing an SSH session, the terminal wouldn't produce the right output because we always emulate the ONLCR flag. Instead, assuming we remove the emulation from the library, we should allocate a PTY and pass the SSH PTY request terminal modes to the allocated PTY to get the correct output (with ONLCR disabled).

[^1]: https://github.com/openssh/openssh-portable/blob/master/clientloop.c#L1350
[^2]: https://github.com/openssh/openssh-portable/blob/master/sshtty.c#L69

belak:
Is working on better PTY support you'd be interested in, or not really? I realize that's a fairly big ask though, especially compared to what you're asking for in this PR.

aymanbagabas:
> Is working on better PTY support you'd be interested in, or not really? I realize that's a fairly big ask though, especially compared to what you're asking for in this PR.

I've implemented PTY support [here](https://github.com/charmbracelet/ssh/pull/8). This is based off https://github.com/gliderlabs/ssh/pull/185 and https://github.com/gliderlabs/ssh/pull/210. It uses [go-pty](https://github.com/aymanbagabas/go-pty).

nichtverstehen:
(Consumer of this package here)

This change seems safe to merge. It adds faithful parsing of the protocol according to the RFC.

It's a useful change that enables callers to observe terminal flags coming on the wire. It leaves the actual handling of PTY to the caller (consistently with the rest of the API of this package).

As a matter of fact, unaware of this PR I implemented the same in a private fork.

Any concerns with merging this?

blaubaer:
Thanks a lot for bringing this work together, @aymanbagabas, and for preserving the credit to @maisem and Tailscale.

We've implemented and extended the terminal-mode support in our maintained fork of [gliderlabs/ssh](https://github.com/gliderlabs/ssh) and merged it here: https://github.com/engity-com/ssh-server-go/pull/7

The implementation exposes complete PTY request metadata, including terminal modes and pixel dimensions, and adds stricter parsing and additional handling for malformed input, opcode boundaries, map isolation, and platform-specific integer conversions.

Both this PR and #131 are explicitly referenced in our PR, including the original credit to Maisem Ali and Tailscale, so that the lineage of the work remains clear.

We're maintaining [engity-com/ssh-server-go](https://github.com/engity-com/ssh-server-go) as a continuation of [gliderlabs/ssh](https://github.com/gliderlabs/ssh), since upstream development has unfortunately become rather quiet. Thanks for the work that went into getting this functionality this far.

PR #211, opened 2023-08-09
Make sshContext thread safe and fix the data race bug

### LINKED ISSUE #117: Data race warning
Hello,

I'm using you awesome library to help testing something that creates a ssh client without relying on something installed on the host that execute the tests.

When testing, I got the following data race:
```
 ==================
 WARNING: DATA RACE
 Write at 0x00c0001680d8 by goroutine 38:
   sync.(*WaitGroup).Wait()
       /usr/local/go/src/internal/race/race.go:41 +0xef
   github.com/gliderlabs/ssh.(*Server).Shutdown.func1()
       /go/pkg/mod/github.com/gliderlabs/ssh@v0.2.2/server.go:178 +0x55

 Previous read at 0x00c0001680d8 by goroutine 25:
   sync.(*WaitGroup).Add()
       /usr/local/go/src/internal/race/race.go:37 +0x169
   github.com/gliderlabs/ssh.(*Server).trackConn()
       /go/pkg/mod/github.com/gliderlabs/ssh@v0.2.2/server.go:389 +0x107
   github.com/gliderlabs/ssh.(*Server).handleConn()
       /go/pkg/mod/github.com/gliderlabs/ssh@v0.2.2/server.go:259 +0x2c7

 Goroutine 38 (running) created at:
   github.com/gliderlabs/ssh.(*Server).Shutdown()
       /go/pkg/mod/github.com/gliderlabs/ssh@v0.2.2/server.go:176 +0xef
   app/pkg/sshx.newSSHServer.func3()
       /app/pkg/sshx/client_test.go:89 +0xd0
   app/pkg/sshx.TestNewClient.func3()
       /app/pkg/sshx/client_test.go:118 +0x39b
   testing.tRunner()
       /usr/local/go/src/testing/testing.go:865 +0x163

 Goroutine 25 (running) created at:
   github.com/gliderlabs/ssh.(*Server).Serve()
       /go/pkg/mod/github.com/gliderlabs/ssh@v0.2.2/server.go:230 +0x184
   app/pkg/sshx.newSSHServer.func2()
       /app/pkg/sshx/client_test.go:77 +0x59
 ==================
```

It happened only one time, and I didn't succeeds to reproduce as restarting the same test without touching anything make it PASS, and so far I didn't get this data race again.

Here is how I call your library:

```go
func TestNewClient(t *testing.T) {
	addr, stop, exited, err := newSSHServer()
	require.NoError(t, err)

	client, err := NewClient(addr) // GITHUB: this returns a *ssh.Client configured to hit the configured ssh server
	require.NoError(t, err)
	assert.Equal(t, "SSH-2.0-go-test_"+addr, string(client.ServerVersion())) // GITHUB: dumb test, just making sure we're connected on the server we wanted
	assert.NoError(t, client.Close())

	stop()
	assert.NoError(t, <-exited)
}

func newSSHServer() (string, func(), <-chan error, error) {
	listener, err := findFreeListener() // GITHUB: this helps to find a listener available
	if err != nil {
		return "", nil, nil, errors.Wrap(err, "unable to get free listener")
	}
	addr := listener.Addr().String()

	srv := gssh.Server{
		Addr:                        addr,
		Version:                     "go-test_" + addr,

		LocalPortForwardingCallback: func(gssh.Context, string, uint32) bool { return true },
		ChannelHandlers: map[string]gssh.ChannelHandler{
			"direct-tcpip": gssh.DirectTCPIPHandler,
		},
	}
	cerr := make(chan error)
	go func() {
		if err := srv.Serve(listener); err != nil {
			if errors.Cause(err) != gssh.ErrServerClosed {
				cerr <- errors.Wrap(err, "unable to serve ssh")
			}
		}
		cerr <- nil
	}()

	return addr, func() {
		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
		defer cancel()

		if err := srv.Shutdown(ctx); err != nil {
			select {
			case cerr <- err:
			default:
			}
		}
	}, cerr, nil
}
```

I'm not sure how critical this can be, nor why I got it only one time, but I wanted to notify you about this.

Thanks for your work on this library,
krostar

### LINKED ISSUE #160: Data race on sshContext

wxiaoguang:
Fix #160

The problem in old code, suppose there are 2 goroutines:

1. Goroutine-A calls the `context.Context` interface functions of `sshCtx`, then it reads `sshCtx.Context`
2. Goroutine-B calls `sshCtx.SetValue`, then it writes `sshCtx.Context`
3. Then, data race occurs

This PR makes the `sshCtx.Context` immutable, there won't be data race issue.

The `Mutex.Lock/Unlock` is quite fast when there is no concurrency, it only calls `atomic.CompareAndSwapInt32` / `atomic.AddInt32`.

### DIFF
--- a/context.go
+++ b/context.go
@@ -94,11 +94,14 @@ type Context interface {
 type sshContext struct {
 	context.Context
 	*sync.Mutex
+
+	values   map[interface{}]interface{}
+	valuesMu sync.Mutex
 }
 
 func newContext(srv *Server) (*sshContext, context.CancelFunc) {
 	innerCtx, cancel := context.WithCancel(context.Background())
-	ctx := &sshContext{innerCtx, &sync.Mutex{}}
+	ctx := &sshContext{Context: innerCtx, Mutex: &sync.Mutex{}, values: make(map[interface{}]interface{})}
 	ctx.SetValue(ContextKeyServer, srv)
 	perms := &Permissions{&gossh.Permissions{}}
 	ctx.SetValue(ContextKeyPermissions, perms)
@@ -119,8 +122,19 @@ func applyConnMetadata(ctx Context, conn gossh.ConnMetadata) {
 	ctx.SetValue(ContextKeyRemoteAddr, conn.RemoteAddr())
 }
 
+func (ctx *sshContext) Value(key interface{}) interface{} {
+	ctx.valuesMu.Lock()
+	defer ctx.valuesMu.Unlock()
+	if v, ok := ctx.values[key]; ok {
+		return v
+	}
+	return ctx.Context.Value(key)
+}
+
 func (ctx *sshContext) SetValue(key, value interface{}) {
-	ctx.Context = context.WithValue(ctx.Context, key, value)
+	ctx.valuesMu.Lock()
+	defer ctx.valuesMu.Unlock()
+	ctx.values[key] = value
 }
 
 func (ctx *sshContext) User() string {
--- a/context_test.go
+++ b/context_test.go
@@ -1,6 +1,9 @@
 package ssh
 
-import "testing"
+import (
+	"testing"
+	"time"
+)
 
 func TestSetPermissions(t *testing.T) {
 	t.Parallel()
@@ -43,5 +46,40 @@ func TestSetValue(t *testing.T) {
 	defer cleanup()
 	if err := session.Run(""); err != nil {
 		t.Fatal(err)
+	}
+}
+
+func TestSetValueConcurrency(t *testing.T) {
+	ctx, cancel := newContext(nil)
+	defer cancel()
+
+	go func() {
+		for { // use a loop to access context.Context functions to make sure they are thread-safe with SetValue
+			_, _ = ctx.Deadline()
+			_ = ctx.Err()
+			_ = ctx.Value("foo")
+			select {
+			case <-ctx.Done():
+				break
+			default:
+			}
+		}
+	}()
+	ctx.SetValue("bar", -1) // a context value which never changes
+	now := time.Now()
+	var cnt int64
+	go func() {
+		for time.Since(now) < 100*time.Millisecond {
+			cnt++
+			ctx.SetValue("foo", cnt) // a context value which changes a lot
+		}
+		cancel()
+	}()
+	<-ctx.Done()
+	if ctx.Value("foo") != cnt {
+		t.Fatal("context.Value(foo) doesn't match latest SetValue")
+	}
+	if ctx.Value("bar") != -1 {
+		t.Fatal("context.Value(bar) doesn't match latest SetValue")
 	}
 }


gustavosbarreto:
@wxiaoguang Could you please improve the commit message?

What is the difference from #209?

silverwind:
I suspect that this race was exposed by some change in the go1.21 compiler. These test failures appeared in the last 1-2 days for gitea and our CI did automatically upgrade to 1.21 when it went live.

context.go
```diff
@@ -119,8 +122,29 @@ func applyConnMetadata(ctx Context, conn gossh.ConnMetadata) {
...
+}
+
+func (ctx *sshContext) Value(key interface{}) interface{} {
+	return ctx.parentContext().Value(key)
+}
+
 func (ctx *sshContext) SetValue(key, value interface{}) {
-	ctx.Context = context.WithValue(ctx.Context, key, value)
+	newCtx := context.WithValue(ctx.parentContext(), key, value)
+	ctx.parentCtx.Store(&newCtx)
```
KN4CK3R, line 147: That's still not safe if A and B both calls `SetValue` at the same time only one value will be stored.

wxiaoguang: Good point, although it is the old behavior, it's not ideal ... Will rewrite it.

wxiaoguang: Done in d9f5b41d17848bfbfc072ec9c3814ec77f3e78a0

KN4CK3R:
Now it's the same as #209 ?

wxiaoguang:
> Now it's the same as #209 ?

No. https://github.com/gliderlabs/ssh/pull/211#issuecomment-1671849486

>>  More details in https://github.com/gliderlabs/ssh/issues/160#issuecomment-1671192825 , IMO `209` is incomplete, because all functions of `context.Context` interface should be thread-safe.

KN4CK3R:
But they are not thread-safe in this PR too.

wxiaoguang:
> But they are not thread-safe in this PR too.

This one is safe, while that one isn't

See the comment https://github.com/gliderlabs/ssh/issues/160#issuecomment-1671192825

> All functions of context.Context interface should be thread-safe.
> Every call to sshCtx.Context (including Err() and others) all causes Data Race.

The struct: ` sshContext { context.Context }`, then, calling `sshCtx.Err()` means `sshCtx.Context.Err()`

* This PR: `sshCtx.Context` is immutable, so it is thread-safe.
* That PR: `sshCtx.Context` is mutable and will be changed by `SetValue`.

wxiaoguang:
And one more thing: changing "ctx" again and again is an anti-pattern of Golang design. So, it's good to make it immutable.

KN4CK3R:
>     * This PR: `sshCtx.Context` is immutable, so it is thread-safe.

Understand.

With the original Go `WithValue` it's not possible to retrieve multiple values added with the same `key`, you always get the last added. So you could use a `map` to get the same behaviour and cleaner code. Currently you always get the oldest value added with the same key.

wxiaoguang:
> Currently you always get the oldest value added with the same key.

Nope, it  (`SetValue`) behaves exactly as `original Go WithValue`, see the "for" loop in `SetValue`. I will add a test for it.

KN4CK3R:
Why use a list then if you replace the previous value? That's the same behaviour as a `map`.

wxiaoguang:
> Why use a list then if you replace the previous value? That's the same behaviour as a `map`.

The reasons IMO are:

1. That's the same behavior as `WithValue`
2. That's the same behavior as the old `SetValue`

KN4CK3R:
>     1. That's the same behavior as `WithValue`

What behaviour are you talking about? The current implementation behaves like a map. The Go `WithValue` keeps a reference to all added values, with the current implementation they may get GCed (not a real world problem I think).

>     3. The context "key" type is `any` which is not comparable (defined by `Value` / `WithValue`), can't be used as a map key, so it can't use a map

`any` is comparable since Go 1.20.

wxiaoguang:
> > 1. That's the same behavior as `WithValue`
> 
> What behaviour are you talking about? The current implementation behaves like a map. The Go `WithValue` keeps a reference to all added values, with the current implementation they may get GCed (not a real world problem I think).

The behavior:

```
ctx = ctx.WithValue(key, v1)
ctx = ctx.WithValue(key, v2)
then ctx.Value(key) == v2, and the Value function is virtually a "recursion loop" of traversing all the keys.
```

```
ctx.SetValue(key, v1)
ctx.SetValue(key, v2)
then ctx.Value(key) == v2, and the sshContext.Value function is real loop of traversing all the keys.
```

For the GC, I do not see a real bad case for it. Go developers should seldom depend on this behavior.

> 
> > 3. The context "key" type is `any` which is not comparable (defined by `Value` / `WithValue`), can't be used as a map key, so it can't use a map
> 
> `any` is comparable since Go 1.20.

For this project:

image

image

KN4CK3R:
> The behavior:

Yes, that's the normal behaviour and would be identical with `map`. I thought you are talking about something else.

> For this project:

That's sad

wxiaoguang:
> > For this project:
> 
> That's sad

Hmm, I think you are right for the map with `interface{}` key. Although Go 1.20 makes the "any" to be "comparable", while the "any/interface{}" could be used as a map key in very old Go releases, the CI passes.

wxiaoguang:
@gustavosbarreto do you have more questions?

gustavosbarreto:
> @gustavosbarreto do you have more questions?

LGTM

gustavosbarreto (approved):

gustavosbarreto:
Can you please squash the commits so we have a clean history without fix commits being applied on top of an original patch.

wxiaoguang:
> Can you please squash the commits so we have a clean history without fix commits being applied on top of an original patch.

Done, while IMO it's also good to do "squash" when the mergers is doing the final merge (into master), then contributors don't need to squash and the commit history is clear for reviewers.

context.go
```diff
@@ -119,8 +122,19 @@ func applyConnMetadata(ctx Context, conn gossh.ConnMetadata) {
 	ctx.SetValue(ContextKeyRemoteAddr, conn.RemoteAddr())
 }
 
+func (ctx *sshContext) Value(key interface{}) interface{} {
```
puni9869, line 125:
```suggestion
func (ctx *sshContext) Value(key any) any {
```

Same at other places

puni9869 (approved):

wxiaoguang:
Kindly ping ~~~

gustavosbarreto:
> Kindly ping ~~~

@belak Can you please take a look into this?

wxiaoguang:
Kindly ping (2 weeks) ~~~

wxiaoguang:
Kindly ping (another 2 weeks) ~~~

I don't like keeping stale PRs. If this PR doesn't get interest or this project is dead, I will close it.

wxiaoguang:
Hmm, another week, no interest. Close and feel free to take over.

gustavosbarreto:
@wxiaoguang I apologize for the delay. I could have merged your PR, but I preferred to wait for other long-time contributors to approve it as well. Since no one has responded to my requests, I will proceed with the merge because I believe this fix is important.

belak:
Sorry about the delay in reviewing - I forgot I set up a filter to put all Github PRs in a folder. Either way, this looks good to me - thanks for submitting it!

PR #241, opened 2024-11-08
Fix ineffective break statement in context_test.go

mfielding:
From https://staticcheck.dev/docs/checks#SA4011:

SA4011 - Break statement with no effect. Did you mean to break out of an outer loop?

In Go, the break statement only exits the innermost loop it's currently in. If you're seeing the "ineffective break statement" warning, it's likely because you have a break statement inside a nested loop, but you intended to exit the outer loop.

Updating with a `return` to exit the goroutine entirely.  The unit test itself isn't affected;  it passes before and after this change.

### DIFF
--- a/context_test.go
+++ b/context_test.go
@@ -60,7 +60,7 @@ func TestSetValueConcurrency(t *testing.T) {
 			_ = ctx.Value("foo")
 			select {
 			case <-ctx.Done():
-				break
+				return
 			default:
 			}
 		}


blaubaer:
Thanks a lot for spotting and fixing this, @mfielding — the ineffective `break` was easy to miss, and your change made the intended goroutine lifecycle much clearer.

We've ported your fix to our maintained fork of [gliderlabs/ssh](https://github.com/gliderlabs/ssh) and incorporated it here: https://github.com/engity-com/ssh-server-go/pull/6

We're maintaining [engity-com/ssh-server-go](https://github.com/engity-com/ssh-server-go) as a continuation of [gliderlabs/ssh](https://github.com/gliderlabs/ssh), since upstream development has unfortunately become rather quiet. Contributions like this are extremely valuable to us, so thank you for taking the time to identify and fix this.

PR #248, opened 2025-04-16
fix: handle Close() before Serve()

kian99:
This change prevents race conditions in programs that call the server's `Close()` or `Shutdown()` methods before `Serve()`. Consider a program that runs both `Serve()` and `Close()` in separate go routines, if a signal comes in early telling the program to shutdown, the `Close()` method may be called before `Serve()` has completed its initialisation and begun tracking the listener.

To fix this I've simply copied the latest approach from the Go standard library HTTP server and applied them. Since the SSH server was based on Go's stdlib implementation, bringing the latest stdlib changes was easy. For reference, we can see that in e.g. [Go 1.10](https://github.com/golang/go/blob/release-branch.go1.10/src/net/http/server.go), there is very little difference between the two. [Go 1.11](https://github.com/golang/go/blob/release-branch.go1.11/src/net/http/server.go#L2904) then removed the logic in `trackListener` that allowed the server to be reused and [Go 1.20](https://github.com/golang/go/blob/release-branch.go1.20/src/net/http/server.go) made several changes like the removal of the `doneChan` in favour of an atomic boolean. 

While technically the change/inability to reuse a previously closed SSH server is a breaking one, the same issue was raised for Go's stdlib [here](https://github.com/golang/go/issues/24047) and [here](https://github.com/golang/go/issues/20239) where it was fixed.

### DIFF
--- a/server.go
+++ b/server.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	gossh "golang.org/x/crypto/ssh"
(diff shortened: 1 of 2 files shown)


kian99:
Hi @gustavosbarreto or @belak, would either of you be able to take a look at this?

PR #259, opened 2026-05-16
feat: Adding support for HA Proxy's PROXY protocol

donovanhubbard:
In 2011 the people working on HA Proxy developed what would be known as the PROXY protocol. This solved the problem of how upstream load balancers and proxies could pass information about the original client's IP address and port to the backend server. 

In order to take advantage of the PROXY protocol, the load balancer must send the PROXY header and the backend server needs to be able to read the header. The protocol is supported by a wide variety of load balancers including:

- HA Proxy
- nginx
- Amazon Application Load Balancers
- Traefik
- Envoy
- Cloudflare

This pull request adds support for the PROXY protocol to the ssh server via a new option called `EnableProxyProtocol`. An example has been included in the `_examples` directory.

This is accomplished through the https://github.com/pires/go-proxyproto library. It replaces the information returned by the `ssh.Session` object's `RemoteAdd()` function with the requestor's original IP address and port and not the load balancer's load balancer and port which is what would be displayed without this protocol.

### DIFF
--- a/_examples/ssh-proxy-protocol/proxy_protocol.go
+++ b/_examples/ssh-proxy-protocol/proxy_protocol.go
@@ -5,7 +5,7 @@ import (
 	"io"
 	"log"
 
-	"github.com/donovanhubbard/ssh"
+	"github.com/gliderlabs/ssh"
 )
 
 const (
--- a/_examples/ssh-proxy-protocol/proxy_protocol.go
+++ b/_examples/ssh-proxy-protocol/proxy_protocol.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"log"
+
+	"github.com/donovanhubbard/ssh"
+)
+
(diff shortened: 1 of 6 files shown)


blaubaer:
We've reimplemented this PR in [engity-com/ssh-server-go#10](https://github.com/engity-com/ssh-server-go/pull/10), with this PR and @donovanhubbard explicitly credited as the origin of the work.

The implementation has been adapted to the current server lifecycle and configuration model and extended with configurable connection policies, header validation, timeouts, support for both Serve and HandleConn, and broader lifecycle and security test coverage.

We're maintaining [engity-com/ssh-server-go](https://github.com/engity-com/ssh-server-go) as a continuation of [gliderlabs/ssh](https://github.com/gliderlabs/ssh), since upstream development has unfortunately become rather quiet. Thanks for putting together the original PROXY protocol implementation.

PR #262, opened 2026-07-19
Fix integer overflow in parseString bounds check

arpitjain099:
I do supply-chain security research and found this while auditing SSH server parsers.

parseString reads a uint32 length prefix and then guards with `uint32(len(in)) < 4+length`. The `4+length` is evaluated in uint32, so a length close to the max wraps around. With `length = 0xFFFFFFFF`, `4+length` becomes 3, the check passes for any buffer of four or more bytes, and execution reaches `in[4 : 4+length]` which is `in[4:3]` and panics with a slice-bounds error.

You can hit this with a malformed pty-req: parsePtyRequest feeds the term field straight into parseString. There is no recover() in the package and DefaultSessionHandler runs in its own goroutine, so the panic brings the whole process down rather than just failing the one connection. When a server sets neither a PublicKeyHandler nor a PasswordHandler, NoClientAuth is enabled, so on those setups an unauthenticated client can trigger it.

The fix advances past the length prefix first and then compares `len(in)` against `length` directly, which is the same shape golang.org/x/crypto/ssh uses and avoids the addition entirely. Valid inputs parse exactly as before.

Added util_test.go: a normal round-trip case, the 0xFFFFFFFF length case, and a malformed pty-req. The oversized cases panic on master and pass with this change; `go test ./...` is green after the fix.


blaubaer:
Thanks a lot for finding, documenting, and fixing this issue, @arpitjain099 — especially for the detailed analysis and regression tests.

We've ported your fix to our maintained fork of [gliderlabs/ssh](https://github.com/gliderlabs/ssh) and merged it here: https://github.com/engity-com/ssh-server-go/pull/4

We're maintaining [engity-com/ssh-server-go](https://github.com/engity-com/ssh-server-go) as a continuation of [gliderlabs/ssh](https://github.com/gliderlabs/ssh), since upstream development has unfortunately become rather quiet. Contributions like this are extremely valuable to us, so thank you for putting the work into investigating and fixing this.

PR #263, opened 2026-08-26
Add disconnect callback

LukaszMoskala:
## Summary
Adds a `DisconnectCallback`, invoked when a connection dies *after* the SSH session was successfully established. This complements the existing `ConnectionFailedCallback`, which only covers connection failures.

## Motivation
With SSH `ControlMaster`, multiple interactive sessions can share a single underlying SSH connection. When those sessions share resources tied to the connection, there needs to be a reliable hook to release them once the connection actually closes.

Releasing resources after the last interactive session closes isn't sufficient: with `ControlMaster`, the underlying SSH connection may still be alive and can still open new interactive sessions. `DisconnectCallback` fires specifically when the SSH connection itself ends, so cleanup only happens when no further sessions can be opened.

I have tested this and it works as expected.

Thank you for maintaining this library and have a nice day!


LukaszMoskala:
Interesting, tests pass on my local machine (`go version go1.27.0-X:nodwarf5 linux/amd64`) every time.

Can this test be re-run?

blaubaer:
Thanks a lot for proposing and implementing this, @LukaszMoskala — the `DisconnectCallback` fills an important gap in the connection lifecycle, especially for cases like SSH `ControlMaster` where resources are tied to the underlying connection rather than individual sessions.

We've ported your change to our maintained fork of [gliderlabs/ssh](https://github.com/gliderlabs/ssh) here: https://github.com/engity-com/ssh-server-go/pull/5

We're maintaining [engity-com/ssh-server-go](https://github.com/engity-com/ssh-server-go) as a continuation of [gliderlabs/ssh](https://github.com/gliderlabs/ssh), since upstream development has unfortunately become rather quiet. Contributions like this are extremely valuable to us, so thank you for putting the work into designing and implementing this.

